PR #7, opened 2020-02-14
Add env variables

ZuGao:
Added the following ENV variables for Apache and ModSecurity configuration:
* TIMEOUT
* LOGLEVEL
* ERRORLOG
* USER
* GROUP
* SERVERADMIN
* SERVERNAME
* PORT
* MODSEC_RULE_ENGINE
* MODSEC_REQ_BODY_ACCESS
* MODSEC_REQ_BODY_LIMIIT
* MODSEC_REQ_BODY_NOFILES_LIMIT
* MODSEC_RESP_BODY_ACCESS
* MODSEC_RESP_BODY_LIMIT
* MODSEC_PCRE_MATCH_LIMIT
* MODSEC_PCRE_MATCH_LIMIT_RECURSION


src/modsecurity.conf
```diff
@@ -0,0 +1,226 @@
...
+# to properly map encoded data to your language. Properly setting
+# these directives helps to reduce false positives and negatives.
+#
+SecUnicodeMapFile unicode.mapping 20127
+
+# Improve the quality of ModSecurity by sharing information about your
+# current ModSecurity version and dependencies versions.
+# The following information will be shared: ModSecurity version,
+# Web Server version, APR version, PCRE version, Lua version, Libxml2
+# version, Anonymous unique id for host.
```
bittner: Also here, maybe a bit less critical but still, we the verbose comments are distracting. I'd suggest to group together related `Sec` directives to maximize readability.

Include a link to the original, unaltered default version of this file in the upstream project, instead!

src/httpd-default.conf
```diff
@@ -0,0 +1,139 @@
...
+# Set a timeout for how long the client may take to send the request header
+# and body.
+# The default for the headers is header=20-40,MinRate=500, which means wait
+# for the first byte of headers for 20 seconds. If some data arrives,
+# increase the timeout corresponding to a data rate of 500 bytes/s, but not
+# above 40 seconds.
+# The default for the request body is body=20,MinRate=500, which is the same
+# but has no upper limit for the timeout.
+# To disable, set to header=0 body=0
+#
```
bittner: (Looks like my earlier comment got lost.)

The excessive use of comments is distracting from the details of the implementation. Please, remove the comments, group related directives together to enhance readability, and include a link to the original, unaltered default version of this file in the upstream repository.

3.3-apache/Dockerfile
```diff
@@ -28,11 +45,11 @@ RUN apt-get update \
  && sed -i /etc/modsecurity.d/modsecurity.conf \
       -e 's/SecRuleEngine DetectionOnly/SecRuleEngine On/g'
 
-COPY httpd-logging-before-modsec.conf /usr/local/apache2/conf/extra/
-COPY httpd-logging-after-modsec.conf  /usr/local/apache2/conf/extra/
```
bittner: I'm not sure whether we'll still need those files.

In [vshn/modsecurity](https://github.com/vshn/modsecurity-docker/blob/master/v3.1/conf-available/logging.conf) we have all the logging configuration in a single configuration file.

3.3-apache/Dockerfile
```diff
@@ -5,10 +5,27 @@ LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
 ARG COMMIT=v3.3/dev
 ARG BRANCH=v3.3/dev
 ARG REPO=SpiderLabs/owasp-modsecurity-crs
-ENV WEBSERVER=Apache
-ENV PARANOIA=1
-ENV ANOMALYIN=5
-ENV ANOMALYOUT=4
+
+ENV WEBSERVER=Apache \
```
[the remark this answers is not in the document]
ZuGao: Updated the structure of the project, deleted the WEBSERVER ENV variable from both images

src/httpd-default.conf
ZuGao: Couldn't find a an official httpd-default.conf files but we can use a [3rd party](https://gist.github.com/TheSunMan/4008088)

README.md
```diff
@@ -25,6 +25,10 @@ or
 docker build -t owasp/modsecurity-crs .
 docker run -p 80:80 -ti -e PARANOIA=4 -e PROXY=1 --rm owasp/modsecurity-crs
 ```
+
+## Apache
+The Apache webserver is configured via the ```httpd-default.conf``` file overriding directives from the base file.
```
mhutter:
```suggestion
The Apache webserver is configured via the `httpd-default.conf` file overriding directives from the base file.
```

3.3-nginx/Dockerfile
```diff
@@ -28,7 +27,8 @@ RUN apt-get update \
  && sed -i /etc/modsecurity.d/modsecurity.conf \
       -e 's/SecRuleEngine DetectionOnly/SecRuleEngine On/g'
 
-COPY docker-entrypoint.sh /
+COPY 3.3-nginx/docker-entrypoint.sh /
+COPY common/modsec-activate-rules.sh /
 
 EXPOSE 80
```
mhutter:
```suggestion
EXPOSE ${PORT}
```

3.3-apache/Dockerfile
```diff
@@ -28,11 +44,12 @@ RUN apt-get update \
...
-COPY httpd-logging-after-modsec.conf  /usr/local/apache2/conf/extra/
-COPY docker-entrypoint.sh /
+COPY common/modsecurity.conf /etc/modsecurity.d/
+COPY 3.3-apache/conf/*.conf /usr/local/apache2/conf/extra/
+COPY 3.3-apache/docker-entrypoint.sh /
+COPY common/modsec-activate-rules.sh /
 
-EXPOSE 80
+EXPOSE ${PORT}
 EXPOSE 443
```
mhutter: Is this required?

3.3-nginx/Dockerfile
ZuGao: Actually I left the configuration for nginx for the next pull request that I'm currently working on. I had to touch the nginx part of the repository because of the restructuring of the project that was suggested by @bittner. For that matter I created a new commit on top of the "Add environment variables" in this pull request.

3.3-apache/Dockerfile
ZuGao: if the port is not hard-coded anymore I would say definitely yes

ZuGao:
> `3.3-apache/Dockerfile` should probably inherit from `owasp/modsecurity:3` as the nginx one does?

In this ticket we are making available environment variables available so I'm not sure if it's the right moment to update the upstream image version. @bittner can you comment on that?

3.3-apache/Dockerfile
```diff
@@ -28,11 +44,12 @@ RUN apt-get update \
...
-COPY httpd-logging-before-modsec.conf /usr/local/apache2/conf/extra/
-COPY httpd-logging-after-modsec.conf  /usr/local/apache2/conf/extra/
-COPY docker-entrypoint.sh /
+COPY common/modsecurity.conf /etc/modsecurity.d/
+COPY 3.3-apache/conf/*.conf /usr/local/apache2/conf/extra/
+COPY 3.3-apache/docker-entrypoint.sh /
+COPY common/modsec-activate-rules.sh /
 
-EXPOSE 80
+EXPOSE ${PORT}
```
srueg: I'm not sure this will work. The EXPOSE directive is basically just metadata of the image and as far as I know it's set at build time. So it'll be always `80` anyway.

srueg: We tested it and it is as expected always `80`, so might as well hard-code it to to cause confusion.

3.3-nginx/docker-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+source /modsec-activate-rules.sh
```
bittner: I'd prefer this script to live in `/opt/`.

README.md
```diff
@@ -25,6 +25,10 @@ or
 docker build -t owasp/modsecurity-crs .
 docker run -p 80:80 -ti -e PARANOIA=4 -e PROXY=1 --rm owasp/modsecurity-crs
 ```
+
+## Apache
+The Apache webserver is configured via the `httpd-default.conf` file overriding directives from the base file.
```
bittner: Using a file called `httpd-default.conf` in `/usr/local/apache2/conf/extra/` is a non-standard pattern. I don't think this is a good idea.

We should make it easy for developers knowing _Apache_ (but not this specific setup) to find configuration details where they expect them. A hint in the README is okay as an idea, but sticking to standard patterns would be superior, IMHO.

Other opinions? @mhutter @srueg

common/modsec-activate-rules.sh
```diff
@@ -83,18 +81,4 @@ if [ ! -z $COMBINED_FILE_SIZES ]; then
...
-      fi
-    fi
-  fi
-elif [ $WEBSERVER = "Nginx" ]; then
-  WEBSERVER_ARGUMENTS=''
-fi
-
-
-exec "$@" $WEBSERVER_ARGUMENTS
+exec "$@"
```
bittner: You won't need this line anymore if this script is called by an entrypoint script. You can safely remove it.

common/modsecurity.conf
```diff
@@ -0,0 +1,57 @@
...
+SecAuditEngine RelevantOnly
+SecAuditLogRelevantStatus "^(?:5|4(?!04))"
+SecAuditLogParts ABIJDEFHZ
+SecAuditLogType Serial
+SecAuditLog /var/log/modsec_audit.log
+
+SecArgumentSeparator &
+SecCookieFormat 0
+SecUnicodeMapFile unicode.mapping 20127
+SecStatusEngine On
```
bittner: I's prefer this file to live in `src/<target-directory>` (i.e. `src/etc/modsecurity.d/`).

Mirroring the directory structure has two advantages:

* It is self-explanatory (aka _"Where does this file live in the target image?"_)
* Only one `COPY` command is needed in the `Dockerfile` to integrate all configuration files in the right place. (No updating needed in the future.)

3.3-apache/Dockerfile
```diff
@@ -5,10 +5,26 @@ LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
...
 ARG REPO=SpiderLabs/owasp-modsecurity-crs
-ENV WEBSERVER=Apache
-ENV PARANOIA=1
-ENV ANOMALYIN=5
-ENV ANOMALYOUT=4
+
+ENV PARANOIA=1 \
+    ANOMALY_INBOUND=5 \
+    ANOMALY_OUTBOUND=4 \
+    TIMEOUT=60 \
```
bittner: `TIMEOUT` is too generic, I fear. This makes the parameter hard to understand.

This is called `APACHE_TIMEOUT` in [our VSHN image](https://github.com/vshn/modsecurity-docker#apache), which explains its function more clearly. Should we call it `PROXY_TIMEOUT` or `WEBSERVER_TIMEOUT`?

If this is hard to abstract, as for the wording, we may stick to `APACHE_TIMEOUT` and related `NGINX_*` variables, if needed. Hopefully that's not necessary.

src/httpd-default.conf
bittner: The official Docker httpd image is built at [github.com/docker-library/httpd](https://github.com/docker-library/httpd/blob/master/2.4/Dockerfile).

The official source of the included conf file is probably [docs/conf/httpd.conf.in](https://github.com/apache/httpd/blob/2.4.x/docs/conf/httpd.conf.in) from the Apache project source code.

3.3-apache/Dockerfile
srueg: I'd vote for `APACHE_TIMEOUT` since I don't think there is an equivalent config for nginx.

https://httpd.apache.org/docs/2.4/mod/core.html#timeout

mhutter:
> > `3.3-apache/Dockerfile` should probably inherit from `owasp/modsecurity:3` as the nginx one does?
> 
> In this ticket we are making available environment variables available so I'm not sure if it's the right moment to update the upstream image version.

Makes sense

3.3-apache/Dockerfile
mhutter: It's hard to abstract because it does SO MANY DIFFERENT THINGS according to the docs...

I'd propose `APACHE_TIMEOUT` and then maybe have a generic `PROXY_TIMEOUT` which sets [`ProxyTimeout`](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxytimeout)?

3.3-apache/Dockerfile
mhutter: It _does_ work, but only at build time. So whatever the `ENV` directive inside the Dockerfile sets `$PORT` to...

bittner:
> `3.3-apache/Dockerfile` should probably inherit from `owasp/modsecurity:3` as the nginx one does?

No. This is correct as it is, because:

* [owasp/modsecurity:2](https://github.com/CRS-support/modsecurity-docker/tree/master/v2-apache) == ModSecurity 2 + Apache (since version 3 doesn't run reliably on Apache) 
* [owasp/modsecurity:3](https://github.com/CRS-support/modsecurity-docker/tree/master/v3-nginx) == ModSecurity 3 + Nginx (since v3 runs faster on Nginx than on Apache)

Think in terms of:

* ModSecurity 2 <==> Apache
* ModSecurity 3 <==> Nginx

Hence, if you want Apache you have to pick ModSecurity 2. That's simply a given.

3.3-apache/Dockerfile
ZuGao: The proxy timeout will be set up in a future update. We have` PROXY_TIMEOUT` directive in `apache-vhosts.conf ` and the equivalent `proxy_connect_timeout` for nginx.

README.md
srueg: As discussed: we'll add our own config file `/usr/local/apache2/conf/extra/httpd-modsecurity.conf` which includes all our config changes (i.e. using env vars). That way we still use the defaults of the base image.

ZuGao:
After some extensive testing I found out Apache does not override the `Listen` directive. I had to use sed in `Dockerfile` in order to override the existing directive in `httpd.conf` file.

src/opt/modsecurity/activate-rules.sh
```diff
@@ -83,18 +81,4 @@ if [ ! -z $COMBINED_FILE_SIZES ]; then
...
-      fi
-    fi
-  fi
-elif [ $WEBSERVER = "Nginx" ]; then
-  WEBSERVER_ARGUMENTS=''
-fi
-
-
-exec "$@" $WEBSERVER_ARGUMENTS
+exec "$@"
```
bittner: This script doesn't need to call a command passed in at all. This line, which originally comes from the entrypoint script, should thus be removed.

dune73:
A few thoughts:

* Please bring back the performance data and the anomaly scores in this or a subsequent PR
* The MODSEC_RESP_BODY_LIMIT=524288 is very low. I suggest to raise it to 1M
* The MODSEC_PCRE_MATCH_LIMIT=1000 is also very low for production. 10K or 100K is more reasonable and still relatively safe
* MODSEC_PCRE_MATCH_LIMIT_RECURSION=1000 is also very low for production. 10K or 100K is more reasonable and still relatively safe

src/opt/modsecurity/activate-rules.sh
```diff
@@ -16,7 +14,7 @@ if [ ! -z $ENFORCE_BODYPROC_URLENCODED ]; then
 fi 
 
 # Inbound and Outbound Anomaly Score
-sed -z -E -i 's/#SecAction.{6}id:900110.*tx\.outbound_anomaly_score_threshold=4\"/SecAction \\\n  \"id:900110, \\\n   phase:1, \\\n   nolog, \\\n   pass, \\\n   t:none, \\\n   setvar:tx.inbound_anomaly_score_threshold='"$ANOMALYIN"',  \\\n   setvar:tx.outbound_anomaly_score_threshold='"$ANOMALYOUT"'\"/' /etc/modsecurity.d/owasp-crs/crs-setup.conf 
+sed -z -E -i 's/#SecAction.{6}id:900110.*tx\.outbound_anomaly_score_threshold=4\"/SecAction \\\n  \"id:900110, \\\n   phase:1, \\\n   nolog, \\\n   pass, \\\n   t:none, \\\n   setvar:tx.inbound_anomaly_score_threshold='"$ANOMALY_INBOUND"',  \\\n   setvar:tx.outbound_anomaly_score_threshold='"$ANOMALY_OUTBOUND"'\"/' /etc/modsecurity.d/owasp-crs/crs-setup.conf
```
bittner: There is a merge conflict with these changes now due to PR #9. Can you please merge in the changes from the main branch or rebase your branch?

It would be worth not to lose your changes, though. PR #9 only fixed the quoting.

PR #11, opened 2020-03-02
Add env variables to nginx

ZuGao:
Added the following env variables to the Nginx image:

* NGINX_KEEPALIVE_TIMEOUT
* ERRORLOG
* LOGLEVEL
* USER
* PORT
* SERVERNAME
* WORKER_CONNECTIONS
* MODSEC_RULE_ENGINE
* MODSEC_REQ_BODY_ACCESS
* MODSEC_REQ_BODY_LIMIT
* MODSEC_REQ_BODY_NOFILES_LIMIT
* MODSEC_RESP_BODY_ACCESS
* MODSEC_RESP_BODY_LIMIT
* MODSEC_PCRE_MATCH_LIMIT
* MODSEC_PCRE_MATCH_LIMIT_RECURSION

Since the Nginx does not recognise the env variables in the conf files, I used `envsubst` in the entrypoint script.

### DIFF
--- a/3.3-nginx/Dockerfile
+++ b/3.3-nginx/Dockerfile
@@ -5,15 +5,38 @@ LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
 ARG COMMIT=v3.3/dev
 ARG BRANCH=v3.3/dev
 ARG REPO=SpiderLabs/owasp-modsecurity-crs
-ENV PARANOIA=1
-ENV ANOMALY_INBOUND=5
-ENV ANOMALY_OUTBOUND=4
+
+ENV PARANOIA=1 \
+    ANOMALY_INBOUND=5 \
+    ANOMALY_OUTBOUND=4 \
+    NGINX_KEEPALIVE_TIMEOUT=60s \
+    ERRORLOG=/var/log/nginx/error.log \
+    LOGLEVEL=warn \
+    USER=nginx \
+    PORT=80 \
+    SERVERNAME=locahost \
+    WORKER_CONNECTIONS=1024 \
+    MODSEC_RULE_ENGINE=on \
+    MODSEC_REQ_BODY_ACCESS=on \
+    MODSEC_REQ_BODY_LIMIT=13107200 \
+    MODSEC_REQ_BODY_NOFILES_LIMIT=131072 \
+    MODSEC_RESP_BODY_ACCESS=on \
+    MODSEC_RESP_BODY_LIMIT=1048576 \
+    MODSEC_PCRE_MATCH_LIMIT=100000 \
+    MODSEC_PCRE_MATCH_LIMIT_RECURSION=100000
+
+COPY 3.3-nginx/docker-entrypoint.sh /
+COPY 3.3-nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf
+COPY 3.3-nginx/nginx.conf /etc/nginx/nginx.conf
+COPY src/opt/modsecurity/activate-rules.sh /opt/modsecurity/
+COPY src/etc/modsecurity.d/*.conf /etc/modsecurity.d/
 
 RUN apt-get update \
  && apt-get -y install \
       ca-certificates \
       git \
       iproute2 \
+      moreutils \
  && mkdir /opt/owasp-crs \
  && cd /opt/owasp-crs \
  && git init \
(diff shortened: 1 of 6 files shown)


3.3-nginx/docker-entrypoint.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+envsubst "$(awk 'BEGIN{for(v in ENVIRON) print "$"v}')" <etc/nginx/nginx.conf | sponge etc/nginx/nginx.conf
```
srueg: Maybe a `for` loop would make sense to iterate over the three files?

src/etc/modsecurity.d/modsecurity-override.conf
```diff
@@ -6,7 +6,6 @@ SecRequestBodyAccess ${MODSEC_REQ_BODY_ACCESS}
 
 SecRequestBodyLimit ${MODSEC_REQ_BODY_LIMIT}
 SecRequestBodyNoFilesLimit ${MODSEC_REQ_BODY_NOFILES_LIMIT}
-SecRequestBodyInMemoryLimit 131072
```
srueg: Why is this removed?

srueg: Seems like this is not used anymore in ModSecurity v3.x [1] and the used value (131072) is the default in v2.x anyways [2]. So, good to remove :+1: 

* [1] https://github.com/SpiderLabs/ModSecurity/issues/1516
* [2] https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#secrequestbodyinmemorylimit

3.3-nginx/Dockerfile
```diff
@@ -5,32 +5,46 @@ LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
...
 ARG BRANCH=v3.3/dev
 ARG REPO=SpiderLabs/owasp-modsecurity-crs
-ENV PARANOIA=1
-ENV ANOMALY_INBOUND=5
-ENV ANOMALY_OUTBOUND=4
+
+ENV PARANOIA=1 \
+    ANOMALY_INBOUND=5 \
+    ANOMALY_OUTBOUND=4 \
+    NGINX_KEEPALIVE_TIMEOUT=60 \
```
srueg: Let's add the `s` suffix (for seconds). Also on 
From the nginx docs:

> It is recommended to always specify a suffix

https://nginx.org/en/docs/syntax.html

3.3-nginx/conf.d/default.conf
```diff
@@ -0,0 +1,14 @@
...
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+    }
+
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+}
```
bittner: Please, configure your IDE properly.

In Codium > Preferences > Settings search for "final" and look for

- [x]  Render Final Newline
- [x]  Insert Final Newline
- [x]  Trim Final Newlines

3.3-nginx/docker-entrypoint.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash -e
 
+for file in "etc/nginx/nginx.conf" "etc/nginx/conf.d/default.conf" "etc/modsecurity.d/modsecurity-override.conf"
+do
+    envsubst "$(awk 'BEGIN{for(v in ENVIRON) print "$"v}')" <$file | sponge $file
```
bittner: This will bite us one day.

Can we make this code less cryptic?

3.3-nginx/nginx.conf
```diff
@@ -0,0 +1,32 @@
...
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+
+    keepalive_timeout  ${NGINX_KEEPALIVE_TIMEOUT};
+
+    include /etc/nginx/conf.d/*.conf;
+}
```
bittner: The indenting/spacing is a bit odd and non-uniform. I'd suggest to use only a single space as a separator, on all lines, in all `.conf` files.

3.3-nginx/Dockerfile
```diff
@@ -5,32 +5,46 @@ LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
...
+COPY 3.3-nginx/nginx.conf /etc/nginx/nginx.conf
+COPY src/opt/modsecurity/activate-rules.sh /opt/modsecurity/
+COPY src/etc/modsecurity.d/*.conf /etc/modsecurity.d/
 
 RUN apt-get update \
  && apt-get -y install \
       ca-certificates \
       git \
       iproute2 \
+      moreutils \
```
bittner: Okay, that's for `sponge`.

3.3-nginx/docker-entrypoint.sh
ZuGao: What do you mean by less cryptic? The previous version was without for loop.

PR #12, opened 2020-07-09
Update crs 3.4 dev

fzipi:
This PR will:
- update repo link on v3.3, and point it to master
- add the new v3.4/dev

### DIFF
--- a/.github/workflows/buildimage.yml
+++ b/.github/workflows/buildimage.yml
@@ -10,7 +10,9 @@ jobs:
       matrix:
         image:
         - 3.3-apache
+        - 3.4-apache
         - 3.3-nginx
+        - 3.4-nginx
     steps:
     - uses: actions/checkout@v1
       with:
--- a/3.3-apache/Dockerfile
+++ b/3.3-apache/Dockerfile
@@ -2,9 +2,9 @@ FROM owasp/modsecurity:2
 
 LABEL maintainer="Chaim Sanders <chaim.sanders@gmail.com>"
 
-ARG COMMIT=v3.3/dev
-ARG BRANCH=v3.3/dev
-ARG REPO=SpiderLabs/owasp-modsecurity-crs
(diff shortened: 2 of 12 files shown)


bittner:
Good that we didn't merge this yet.

The original idea was that we have a single source of truth for the `Dockerfile` and its related configuration, which should ideally live in the [`src` folder](https://github.com/coreruleset/modsecurity-crs-docker/tree/master/src), and there should be a `Makefile` (or something similar) to (re)generate the `Dockerfile`s for all the currently supported CRS versions and web proxy combinations.

This was originally described in https://github.com/coreruleset/modsecurity-docker/issues/32 and related issues. Looks like we missed out on implementing this to the full extend, as originally planned.

What Next?
---------

We could

* either merge this PR "as is", and attempt the scripted implementation of generating all `Dockerfile` setups as a follow-up step
* or you could try to implement the scripted approach with this PR directly, before merging it into `master`.

Opinions?

(Personally, I'd prefer to add another commit with the scripted generation to this PR, already. This would make sure the job gets done more likely, I believe. Just a personal opinion, though.)

bittner:
@fzipi Do you want to give the generalization a try?

I'd suggest we do this like in our [appuio/container-oc](https://github.com/appuio/container-oc) repository. There we have a `src` folder and version folders that are generated from `src` with a shell script that mainly does some `sed` magic. The `Makefile` in the root folder is for convenience.

fzipi:
@bittner Sure. Do you think we should add it here, or create a new PR and close this one?

bittner:
> Do you think we should add it here, or create a new PR and close this one?

I think it's fine to replace the commits in this PR. But if you want to keep this PR for future reference go forward, close it and start a new PR. Fine with me.

fzipi:
Need to make the change here by replacing the PR.

fzipi:
Fixed in #25. Closing this one.

PR #20, opened 2020-12-29
fix(vars): use access log variable in nginx container also

fzipi:

ACCESS_LOG was only being used in apache. Let's use it also in nginx.


fzipi:
Nevermind, this is set in the base container :facepalm: . Closing.

PR #21, opened 2020-12-29
feat(utf8): add variable for validating utf8

fzipi:

This variable is being used by the CRS testing framework (taken from the regression test docs):

Requirements
============
There are Three requirements for running the OWASP CRS regressions.

1. You must have ModSecurity specify the location of your error.log, this is done in the config.py file
2. ModSecurity must be in DetectionOnly (or anomaly scoring) mode
3. You must disable IP blocking based on previous events

Note: The test suite compares timezones -- if your test machine and your host machine are in different timezones this can cause bad results

To accomplish 2. and 3. you may use the following rule in your setup.conf:

```
SecAction "id:900005,\
  phase:1,\
  nolog,\
  pass,\
  ctl:ruleEngine=DetectionOnly,\
  ctl:ruleRemoveById=910000,\
  setvar:tx.paranoia_level=4,\
  setvar:tx.crs_validate_utf8_encoding=1,\
  setvar:tx.arg_name_length=100,\
  setvar:tx.arg_length=400"
```

So we need that variable if we want to run proper tests as the doc says.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ The following environment variables are available to configure the CRS container
 | UPSTREAM | The IP Address (and optional port) of the upstream server when proxy mode is enabled. (Default: the container's default router, port 81) (Examples: 192.0.2.2 or 192.0.2.2:80) |
 | EXECUTING_PARANOIA | An integer indicating the executing_paranoia_level (Default: paranoia level) |
 | ENFORCE_BODYPROC_URLENCODED | A boolean indicating the enforce_bodyproc_urlencoded (Default: 0) |
+| VALIDATE_UTF8_ENCODING | A boolean indicating the crs_validate_utf8_encoding (Default: 0) |
 | ANOMALY_INBOUND | An integer indicating the inbound_anomaly_score_threshold (Default: 5) |
 | ANOMALY_OUTBOUND | An integer indicating the outbound_anomaly_score_threshold (Default: 4) |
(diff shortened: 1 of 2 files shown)


fzipi:
After reviewing, we use more parameters for the test that aren't reflected as env vars.

I think the main cause for not taking a look at these is because, well, those tests are disabled.

franbuehler (changes requested):
There is an error in the sed statement and the rule is not set inside the container.

It should be:
```
# Activate UTF8 validation
if [ -n "$VALIDATE_UTF8_ENCODING" ]; then
  sed -z -E -i 's/#SecAction.{6}id:900950.*tx\.crs_validate_utf8_encoding=1\"/SecAction \\\n  \"id:900950, \\\n   phase:1, \\\n   nolog, \\\n   pass, \\\n   t:none, \\\n   setvar:tx.crs_validate_utf8_encoding=1\"/' /etc/modsecurity.d/owasp-crs/crs-setup.conf
fi
```

What I mean is `tx\.crs_validate_utf8_encoding=1` instead of `tx\.setvar:tx.crs_validate_utf8_encoding=1`

Then it works!

fzipi:
@franbuehler Definitely, good catch! Now it should be good to go.

franbuehler (approved):

franbuehler:
LGTM. I did not test it again. But it should work. I tested in the first review.

Issue #18, opened 2020-09-08
Wrong GitHub URl on hub.docker.com


Eshaan7:
The repository on Docker hub points to the wrong/old GitHub URL. https://hub.docker.com/r/owasp/modsecurity-crs/

bittner:
Thanks for reporting this!

The reason why the README on Docker Hub  doesn't reflect the latest advancements is that we build the images here, with GitHub Actions, which doesn't automatically update the Docker Hub README (while builds on Docker Hub do).

We should always push the content of the latest README to Docker Hub when we publish an image. This needs to be done by an API call (see [this SO answer](https://stackoverflow.com/a/57671247/202834) and the [related source code](https://hub.docker.com/r/peterevans/dockerhub-description) for reference).

fzipi:
I tried updating and using the (action from peterevans](https://github.com/peter-evans/dockerhub-description), but the problem is that is need my user and password (doesn't work with tokens), so for now the update must be run manually using:

```
❯ docker run -v $PWD:/workspace \
  -e DOCKERHUB_USERNAME='<user>' \
  -e DOCKERHUB_PASSWORD='<password> \
  -e DOCKERHUB_REPOSITORY='owasp/modsecurity-crs' \
  -e README_FILEPATH='/workspace/README.md' \
  peterevans/dockerhub-description:2
Unable to find image 'peterevans/dockerhub-description:2' locally
2: Pulling from peterevans/dockerhub-description
0a6724ff3fcd: Pull complete 
5fd2bdfdbf4b: Pull complete 
80b224d472a8: Pull complete 
e21405c347ae: Pull complete 
f701b9a7eff1: Pull complete 
a600792de759: Pull complete 
193397fa322a: Pull complete 
Digest: sha256:d472ef009c842a1070459099eb7271af1393fc4cd9cf7aa77231aeabf921f1fb
Status: Downloaded newer image for peterevans/dockerhub-description:2
Acquiring token
Sending PATCH request
Request successful
```

You can see more information about it here: https://github.com/marketplace/actions/update-container-description-action#limitations

fzipi:
Fixed.

PR #23, opened 2021-01-31
fix(docs): add github action to update readme on dockerhub

fzipi:

Adds action to push readme to docker hub.

Solves #18.


fzipi:
@franbuehler Can I have your review here also?

PR #64, opened 2022-03-10
feat(crs): support for new CRS v4

fzipi:

Add new variables in preparation for CRS v4.


fzipi:
Can someone take a look here and see if I'm not breaking/forgetting anything for the CRS v4 new variables naming?

@theseion @theMiddleBlue @franbuehler

docker-compose.yaml
```diff
@@ -31,6 +31,10 @@ services:
       - ANOMALY_OUTBOUND=4
       # Executing Paranoia Level
       # - EXECUTING_PARANOIA=2
+      #
+      # New variable naming
```
theseion, line 35:
```suggestion
      # Replaces PARANOIA as of CRS 4
```

docker-compose.yaml
```diff
@@ -31,6 +31,10 @@ services:
...
+      - BLOCKING_PARANOIA_LEVEL=1
+      # - DETECTION_PARANOIA_LEVEL=4
```
theseion, line 37:
```suggestion
      # Replaces EXECUTING_PARANOIA_LEVEL as of CRS 4
      # - DETECTION_PARANOIA_LEVEL=4
```

src/opt/modsecurity/activate-rules.sh
```diff
@@ -97,7 +107,12 @@ fi
 
 # Substitute MODSEC_TAG
 if [ -n "${MODSEC_TAG}" ]; then
-  sed -z -E -i "s/\\$\{MODSEC_TAG\}/${MODSEC_TAG}/" /etc/modsecurity.d/owasp-crs/crs-setup.conf
+  sed -z -E -i "s/\\$\{MODSEC_TAG\}/${MODSEC_TAG}/g" /etc/modsecurity.d/owasp-crs/crs-setup.conf
+fi
+
+# Reporting Level
+if [ -n "${REPORTING_LEVEL}" ]; then
+  sed -z -E -i 's/#SecAction.{6}id:900115.*tx\.reporting_level=2=1\"/SecAction \\\n  \"id:900115, \\\n   phase:1, \\\n   nolog, \\\n   pass, \\\n   t:none, \\\n   setvar:tx.reporting_level='"${REPORTING_LEVEL}"'\"/' /etc/modsecurity.d/owasp-crs/crs-setup.conf
```
theseion, line 115:
```suggestion
  sed -z -E -i 's/#SecAction.{6}id:900115.*tx\.reporting_level=2\"/SecAction \\\n  \"id:900115, \\\n   phase:1, \\\n   nolog, \\\n   pass, \\\n   t:none, \\\n   setvar:tx.reporting_level='"${REPORTING_LEVEL}"'\"/' /etc/modsecurity.d/owasp-crs/crs-setup.conf
```

docker-compose.yaml
```diff
@@ -31,6 +31,10 @@ services:
...
```
theseion, line 38: Does it make sense to add reporting level here too?

theseion (changes requested):

docker-compose.yaml
fzipi360: Added.

docker-compose.yaml
fzipi: Yes, added.

fzipi:
Thanks @theseion ! Added your suggestions and also described the new variables in the readme.

theseion (commented):
LGTM.

What do you think of fully replacing PARANOIA and EXECUTING_PARANOIA? Technically, we don't need the duplication and for users it makes it more complicated, especially if they switch between CRS < 4 and >= 4.
We could document, how the variables relate to the old naming convention, that should be enough.

fzipi:
Merging.

fzipi:
I agree but for now, let's get this rolling and unblock Simon. Pretty sure we will need more changes until we get to a safe state.

We don't want to break other people's flows also. If we remove the variables, 99% sure that many will suffer the change.

PR #66, opened 2022-03-22
Added includes needed for plugins

theseion:


fzipi:
@theseion Does this work locally?

theseion:
Not yet. Just discovered that I need to create that directory first.

fzipi (approved):
Looks good. We need to think about how plugins will get inside the container. Mounting perhaps?

theseion:
At least for the tests: yes. That's what https://github.com/coreruleset/coreruleset/pull/2448 is for.

fzipi (commented):
After taking a look at https://coreruleset.org/docs/configuring/plugins/#how-to-install-a-plugin, do we want to create 
```
crs/plugins/empty-config.conf
crs/plugins/empty-before.conf
crs/plugins/empty-after.conf
```
?

theseion:
I had the same idea but it doesn't work (at least not by itself). As soon as you mount a plugins directory you have the same issue. Hence, we'd have to add dummy files to the CRS repo and I didn't want to do that. That would only make sense, IMO, if we supplied a prepared configuration file with the `Include ...` lines in it. But since we don't, I'm not sure we should.

theseion:
I need this for https://github.com/coreruleset/coreruleset/issues/2394. So it would be good if we could come to a conclusion.

fzipi (changes requested):
I think this will fail if:
- there is an `/opt/owasp-crs/this-nice-plugins/my-plugin-config.conf` file
- no corresponding symlink in the `/opt/owasp-crs/plugins` directory

If there is a plugin subdirectory infrastructure, e.g: `/opt/owasp-crs/XYZ-plugin/xyz-config,after,before,config}.conf`,  do we recommend symlinking from the `/opt/owasp-crs/plugins/xyz-*.conf` files?

The crs blog post says also:
```
Future CRS releases will come with a plugins folder next to the rules folder. If you do not have that yet then create it and place three empty config files in it (Shell command touch is your friend):

crs/plugins/empty-config.conf
crs/plugins/empty-before.conf
crs/plugins/empty-after.conf

These empty rule files make sure that the web server does not fail when including *.conf if there are no plugins present. (We're aware that Apache supports the IncludeOptional directive, but that is not available on all web servers, so we prefer to use Include for documentation purposes.)
```

Also, plugins directory might not exist, and be volume mounted read-only in docker. So writing by "activating" plugins might fail. 🤔

theseion:
Well spotted! I fixed the `find` operation to look a the full path. I also saw that there was an issue with the second `sed` command (requires `-E`).

The code otherwise accounts for the possibility of the plugins directory being absent (the inlcudes are commented by default and `find` will not fail).

fzipi (approved):
Good, let's see how this works.

theseion:
I don't have write access to the repo, so you'll have to merge.

PR #68, opened 2022-03-28
feat(gpg): add release signature verification

fzipi:

- adds gpg release verification
- splits build in release get and config files copying

### DIFF
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -1,8 +1,26 @@
+FROM owasp/modsecurity:nginx as release
+
+ARG RELEASE
+
+# hadolint ignore=DL3008,SC2016
+RUN set -eux; \
+    apt-get update; \
+    apt-get -y install --no-install-recommends \
+      ca-certificates \
+      curl \
+      gnupg; \
+    mkdir /opt/owasp-crs; \
+    curl -SL https://github.com/coreruleset/coreruleset/archive/v${RELEASE}.tar.gz -o v${RELEASE}.tar.gz; \
+    curl -SL https://github.com/coreruleset/coreruleset/releases/download/v${RELEASE}/coreruleset-${RELEASE}.tar.gz.asc -o coreruleset-${RELEASE}.tar.gz.asc; \
+    gpg --fetch-key https://coreruleset.org/security.asc; \
+    gpg --verify coreruleset-${RELEASE}.tar.gz.asc v${RELEASE}.tar.gz; \
+    tar -zxf v${RELEASE}.tar.gz --strip-components=1 -C /opt/owasp-crs; \
+    rm -f v${RELEASE}.tar.gz coreruleset-${RELEASE}.tar.gz.asc; \
+    mv -v /opt/owasp-crs/crs-setup.conf.example /opt/owasp-crs/crs-setup.conf
+
 FROM owasp/modsecurity:nginx
 
 LABEL maintainer="Felipe Zipitria <felipe.zipitria@owasp.org>"
-
-ARG RELEASE
 
 ENV PARANOIA=1 \
     ANOMALY_INBOUND=5 \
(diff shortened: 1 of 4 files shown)


nginx/Dockerfile
```diff
@@ -30,22 +48,9 @@ COPY src/opt/modsecurity/activate-rules.sh /opt/modsecurity/
 COPY src/etc/modsecurity.d/*.conf /etc/modsecurity.d/
 COPY nginx/docker-entrypoint.sh /
 COPY nginx/nginx.conf /etc/nginx/nginx.conf
+COPY --from=release /opt/owasp-crs /opt/owasp-crs
 
-# Change default shell to bash
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
theseion, line 35: I see this wasn't even in the NGiNX file. Did this even serve any purpose?

nginx/Dockerfile-alpine
```diff
@@ -30,18 +48,14 @@ COPY src/opt/modsecurity/activate-rules.sh /opt/modsecurity/
 COPY src/etc/modsecurity.d/*.conf /etc/modsecurity.d/
 COPY nginx/docker-entrypoint.sh /
 COPY nginx/nginx.conf /etc/nginx/nginx.conf
+COPY --from=release /opt/owasp-crs /opt/owasp-crs
 
-# hadolint ignore=DL3008,SC2016
+# Alpine needs sed because busybox does not support 'z' param
```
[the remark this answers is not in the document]
fzipi: Busybox. Alpine has busybox for basic stuff, that's why you need "regular" sed

nginx/Dockerfile
fzipi: The same that `set -eux` serves now: to fail whenever any command in the list fails.

nginx/Dockerfile-alpine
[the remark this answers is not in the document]
fzipi: See error here: https://github.com/coreruleset/modsecurity-crs-docker/runs/5722658259?check_suite_focus=true

theseion, line 53:
```suggestion
# Alpine needs GNU 'sed' because the 'sed' program shipped with busybox does not support 'z' parameter for separating lines with a 'NUL' character.
```

theseion (changes requested):

fzipi:
@theseion Can we merge this one now?

nginx/Dockerfile-alpine
theseion:
```suggestion
# Alpine needs GNU 'sed' because the 'sed' program shipped with busybox does not support 'z' parameter for separating lines with a 'NUL' character.
```

nginx/Dockerfile-alpine
```diff
@@ -30,18 +48,14 @@ COPY src/opt/modsecurity/activate-rules.sh /opt/modsecurity/
...
 RUN set -eux; \
     apk add --no-cache \
-    ca-certificates \
-    curl \
-    sed; \
-    mkdir /opt/owasp-crs; \
-    curl -SL https://github.com/coreruleset/coreruleset/archive/v${RELEASE}.tar.gz \
-    | tar -zxf - --strip-components=1 -C /opt/owasp-crs; \
-    mv -v /opt/owasp-crs/crs-setup.conf.example /opt/owasp-crs/crs-setup.conf; \
+        sed; \
```
theseion, line 56:
```suggestion
    sed; \
```

theseion (changes requested):
Not sure whether you intentionally ignored that change request, so I'm adding it again.

nginx/Dockerfile-alpine
fzipi: Weird, if you take a look at the files, this change was applied... 🤔

theseion: Oh... sorry. Maybe I was looking at an older commit somehow...

PR #73, opened 2022-04-16
Adapted Dockerfiles and resources to use base image setup

### LINKED ISSUE #72: Modsecurity CRS is expecting a missing file
Starting from yesterday, `owasp/modsecurity-crs:nginx-alpine` refuses to start when using a simple Nginx configuration such as: 
```
modsecurity on;
modsecurity_rules_file /etc/modsecurity.d/setup.conf;
```
The reason seams to be that `/etc/modsecurity.d/setup.conf` is expecting a file named `/etc/modsecurity.d/modsecurity-override.conf` that is missing.

theseion:
Fixes #72.


fzipi (approved):
Ugh, this double imaging is killing use. Thanks for fixing!

PR #75, opened 2022-04-20
Allow nginx resolver directive to be configured

### LINKED ISSUE #74: Unable to configure modsecurity as a reverse proxy
Hello,

I'm trying to add modsecurity into a docker-compose stack.
The modsecurity (waf) is configured as a reverse proxy and should forward requests to a container named `server` 

```
waf:
  image: owasp/modsecurity-crs:3.3.2-nginx
  ports:
    - 80:80
  depends_on:
    - server
  environment:
    - SERVERNAME=localhost
    - PARANOIA=1
    - ANOMALY_INBOUND=5
    - ANOMALY_OUTBOUND=4
    - DNS_SERVER=127.0.0.11
    - PROXY=1
    - BACKEND=http://server/

server:...
```

Unfortunately even with DNS_SERVER provided, modsecurity is unable to resolve the upstream (ie. server). 

The modifications made by the entrypoint named `91-update-resolver.sh` seems to be erased by something (may be another entrypoint from `modsecurity-crs`). So the `resolver` directive is missing in `/etc/nginx/nginx.conf`

As a workaround, i've added a new entrypoint script 

```
  waf:
    image: owasp/modsecurity-crs:3.3.2-nginx
     ...
    volumes:
      - ./waf/99-update-resolver.sh:/docker-entrypoint.d/99-update-resolver.sh:ro
```

Entrypoint script:
```sh
#!/bin/sh
# vim:sw=2:ts=2:sts=2:et

set -eu

DNS_SERVER="${DNS_SERVER:-$(grep -i '^nameserver' /etc/resolv.conf | head -n1 | cut -d ' ' -f2)}"
RESOLVER_DIRECTIVE="resolver ${DNS_SERVER} ipv6=off valid=5s;"

sed -i.bak "/http {/a ${RESOLVER_DIRECTIVE}" /etc/nginx/nginx.conf
```

bguerout:
Add `resolver` directive in `nginx.conf.template` to allow `91-update-resolver.sh` entrypoint from `modsecurity-docker` to patch value

fix #74


fzipi:
Hey @bguerout! Taking a second look at the files... maybe we just need to remove the `nginx.conf.template` file. That way we use everything from upstream, and the DNS_SERVER is already there.

bguerout:
The two templates are not exactly the same but if differences are not required, using the file from upstream is the better solution

fzipi:
I would say update this PR to remove the file in this repo. We don't need the differences.

README.md
```diff
@@ -83,6 +83,7 @@ The following environment variables are available to configure the CRS container
 | APACHE_TIMEOUT | Apache integer value indicating the number of seconds before receiving and sending time out (Default: 60) |
 | NGINX_KEEPALIVE_TIMEOUT | Nginx integer value indicating the number of seconds a keep-alive client connection will stay open on the server side (Default: 60) |
 | LOGLEVEL | A string value controlling the number of messages logged to the error_log (Default: warn) |
+| DNS_SERVER | A string indicating the name servers used to resolve names of upstream servers into addresses. For localhost backend this value should not be defined (Default: not defined) |
```
fzipi, line 86: It is confusing, but this one should come from the base container....

theseion:
Made obsolete by #79.

PR #79, opened 2022-05-29
fix: adapt setup to upstream

### LINKED ISSUE #77: User mismatch in README and dockerfile in Apache
There is a mismatch in what we say in the README and the actual Dockerfile for the Apache build.

We state that the default user is `daemon`, but in fact it is running as `www-data`.

We need to fix this and move to `daemon`, or correct the README.

theseion:
- Remove duplicated, incomplete list of variables and link to
  modesecurity-docker instead
- Use TIMEOUT environment variable from upstream
- Clean up Dockerfiles
- Remove nginx.conf.template to avoid unintened override of upstream
  file

Fixes #77, #75


README.md
```diff
@@ -56,49 +69,44 @@ The Apache webserver is configured via the `httpd-modsecurity.conf` file overrid
...
-| MODSEC_PCRE_MATCH_LIMIT_RECURSION | ModSecurity integer value indicating the limit for the depth of recursion when calling PCRE function (Default: 1000) |
-| MODSEC_DEFAULT_PHASE1_ACTION | ModSecurity string with the contents for the default action in phase 1 (Default: `'phase:1,log,auditlog,pass,tag:\'\${MODSEC_TAG}\''`) |
-| MODSEC_DEFAULT_PHASE2_ACTION | ModSecurity string with the contents for the default action in phase 2 (Default: `'phase:2,log,auditlog,pass,tag:\'\${MODSEC_TAG}\''`) |
-| CRS_ENABLE_TEST_MARKER | A boolean indicating whether to write test markers to the log file (Used for running the CRS test suite. Default: 0) |
+
+### CRS specific
+
+| Name     | Description|
+| -------- | ------------------------------------------------------------------- |
+| PARANOIA | An integer indicating the paranoia level (Default: `1`)               |
```
fzipi, line 88: Shall we sort this list?

theseion: Maybe partially. I would still want things like `*PARANOIA*` to be grouped together, and `PARANOIA` should come before the other `*PARANOIA*` variables because they depend on it.

theseion: I quickly tried sorting but it becomes clear quickly that multiple variables should be grouped together thematically, rather than alphabetically. Having those in the same table doesn't work well. How about one table per group?

fzipi: Probably it makes sense then to keep it simple. Leave it as it is now I guess?

fzipi (approved):
LGTM

PR #80, opened 2022-06-18
fix(apache): use SERVER_ variables from base image

fzipi:

SERVER_NAME and SERVER_ADMIN were used inconsistently in this image and the base. Fix depends on https://github.com/coreruleset/modsecurity-docker/pull/123.


theseion (approved):
LGTM in conjuction with #123

PR #81, opened 2022-06-20
Fix missing ENTRYPOINT in Apache flavor

woefe:
The latest build on dockerhub of the Apache image does not have an ENTRYPOINT directive [1]. 
That leads to `activate-rules.sh` not being sourced and env variables like `ALLOWED_METHODS` not being written to `crs-setup.conf` and therefore not applied. (I noticed ALLOWED_METHOD related messages in my logs a couple of days ago)

Not sure which (parent) image set the ENTRYPOINT before. Hence I introduce it here and not in the parent image to avoid that problem in the future.

[1] https://hub.docker.com/layers/modsecurity-crs/owasp/modsecurity-crs/3.3.2-apache/images/sha256-56461eaca2a6120d10a1ec8406b76be72cdc1ce2db2ba17503f9c820cfe0912c?context=explore

### DIFF
--- a/apache/Dockerfile
+++ b/apache/Dockerfile
@@ -49,4 +49,6 @@ COPY --from=release /opt/owasp-crs /opt/owasp-crs
 
 RUN set -eux; ln -s /opt/owasp-crs /etc/modsecurity.d/
 
+ENTRYPOINT ["/docker-entrypoint.sh"]
+
 # Use httpd-foreground from upstream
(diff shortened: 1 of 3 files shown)


fzipi:
Wow, good find here. I will try to track when this went away! 🤦 

You will need to do a couple of additional changes:
```
index 0d07e2d..b1c1fae 100755
--- a/apache/docker-entrypoint.sh
+++ b/apache/docker-entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/sh -e

 source /opt/modsecurity/activate-rules.sh
```
and add gnu sed to the alpine image like [here](https://github.com/coreruleset/modsecurity-crs-docker/blob/83c94c5e0476dc7a07ffa7115938781d6fd57640/nginx/Dockerfile-alpine#L45-49)

fzipi (approved):
Awesome, thanks!

Let wait for the pipeline, but LGTM.

Thanks for catching this one!

PR #88, opened 2022-09-16
feat(ci): use unique tags when pushing to hub

### LINKED ISSUE #67: Look on how to retain older images in case latest is broken
From #62 , we should investigate if we can have old images laying around in hub.
```
Unfortunately, all images on docker-hub (see: https://hub.docker.com/r/owasp/modsecurity-crs/tags) have just been overridden and it is not possible to pull an older variant. This should definitely be fixed to allow (automatic) rollbacks in case a newer image is broken.
```

fzipi:

- tags are based on build date using YYYYMMDDHHMM

This is the same as https://github.com/coreruleset/modsecurity-docker/pull/140 for this repository.

Fixes #67.

### DIFF
--- a/.github/workflows/buildimage.yml
+++ b/.github/workflows/buildimage.yml
@@ -3,49 +3,61 @@ on:
   push:
     branches:
       - master
-  schedule:
-    - cron: "0 12 * * TUE" # Run every Tuesday (base run on Monday)
   repository_dispatch:
     types: [base-container-updated]
+
+env:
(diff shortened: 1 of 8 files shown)


fzipi:
I'm delaying merging this so we can push a couple changes in the base container. Will use full version in the FROM directives in the build process now.

theseion (approved):
LGTM

theseion (changes requested):

PR #92, opened 2022-10-03
Fix action order in marker rule

theseion:


src/opt/modsecurity/activate-rules.sh
```diff
@@ -132,9 +132,9 @@ if [ -n "${CRS_ENABLE_TEST_MARKER}" ] && [ ${CRS_ENABLE_TEST_MARKER} -eq 1 ] &&
 SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \\
   "id:999999,\\
```
airween, line 133: Can you add two more spaces to each line (increment the indent), as CRS rules [are](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md#general-formatting-guidelines-for-rules-contributions)?

I would like to add crs-setup.conf to the `rules-check.py` too, and without that, the indentation check will show this as error.

airween (changes requested):

src/opt/modsecurity/activate-rules.sh
theseion: done

airween (approved):
Thanks. LGTM.

PR #99, opened 2022-11-11
feat(plugins): enable plugins if we have files

fzipi:

Plugins were enabled automatically if using nginx, but not when using apache. This PR treats every platform in the same way.


theMiddleBlue:
I see the point with the `activate-rules.sh` script, but TBH it's not really clear to me what is the issue with Apache

fzipi:
This is failing only on the nginx build. I have a new branch that will merge this container with the base one, then this will be solved.

theseion (approved):
LGTM.
I wonder whether we should use a separate script though, since it's not really about "activating rules". But that can wait.

fzipi:
Of course, this can be called `activate-plugins.sh`.

fzipi:
Thanks for the fix. I'm rebuilding all this in the new repo merge, so let's push this for now.

PR #111, opened 2023-01-11
Use libpcre3 instead of libpcre++ to reduce build time and image size.

khiemdoan:
ModSecurity depends on libpcre3-dev. libpcre++-dev also depends on libpcre3-dev.
So we can just use libpcre3-dev to avoid redundancy.


fzipi:
Hi @khiemdoan, thanks for the PR.

PR #115, opened 2023-02-05
Release new image

### LINKED ISSUE #114: Re-Build with httpd 2.4.55
Please re-build the official images so that they are based on httpd 2.4.55 - that update fixes 3 CVEs, all with severity medium.  

https://httpd.apache.org/security/vulnerabilities_24.html

fzipi:
- new modsecurity 2.9.7
- use new apache in build

Hopefully fixes #114.


nginx/Dockerfile-alpine
```diff
@@ -4,7 +4,7 @@ FROM nginx:${NGINX_VERSION}-alpine as build
 
 ARG MODSEC_VERSION=3.0.8
 
-# Note: pcre-dev (PCRE3) is required by the build description,
+# Note: pcre-dev (PCRE) is required by the build description,
```
theseion, line 7: That's PCRE version 3, isn't it? The old version.
```suggestion
# Note: pcre-dev (PCRE 3) is required by the build description,
```

theseion (changes requested):
Would it make sense to include the patch version in the image reference? That would make it a bit simpler to do explicit changes (would also be visible in a PR), e.g., `httpd:2.4.55`.

nginx/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM nginx:${NGINX_VERSION} as build
 ARG MODSEC_VERSION=3.0.8 \
     LMDB_VERSION=0.9.29
 
-# Note: libpcre++-dev (PCRE3) is required by the build description,
+# Note: libpcre3-dev (PCRE) is required by the build description,
```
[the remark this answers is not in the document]
khiemdoan: We don't have version 3 yet.
PCRE is name of library.
libpcre3 is name of Debian-based package.

For more infomation:
- https://www.pcre.org/
- https://packages.debian.org/bullseye/libpcre3-dev
- https://en.wikipedia.org/wiki/Perl_Compatible_Regular_Expressions

nginx/Dockerfile-alpine
khiemdoan: We don't have version 3 yet.
PCRE is name of library.
libpcre3 is name of Debian-based package.

For more infomation:
- https://www.pcre.org/
- https://packages.debian.org/bullseye/libpcre3-dev
- https://en.wikipedia.org/wiki/Perl_Compatible_Regular_Expressions

theseion: I'm aware of that. Version 3 of PCRE is the old, obsolete version. Version 2 is the current version. `libpcre3` is version 3. The note is important because we don't actually build with version 3 anymore but it is still required for the build script.

khiemdoan: PCRE doesn't have version 3 now. `libpcre3` is the name of the package, which contains PCRE1.

theseion: You're right. The old version is indeed called PCRE1. The `3` is most likely the ABI version of `libpcre`. I did not know that. Thanks!

theseion: I've updated my review comments to say `PCRE 1`.

fzipi:
> Would it make sense to include the patch version in the image reference? That would make it a bit simpler to do explicit changes (would also be visible in a PR), e.g., `httpd:2.4.55`.

We can, but let's do it for the next release?

nginx/Dockerfile-alpine
theseion, line 7:
```suggestion
# Note: pcre-dev (PCRE 1) is required by the build description,
```

theseion (changes requested):

theseion (approved):

PR #119, opened 2023-02-21
Disable matching of test request

theseion:
Partial fix for https://github.com/coreruleset/coreruleset/pull/3132


dune73:
Looks good.

PR #127, opened 2023-02-26
feat: add apache / nginx variants to docker-compose

theseion:


fzipi (changes requested):
Maybe using the `extends` so we DRY on all the CRS specific variables?

theseion:
Great idea! I did not know about `extends`.

fzipi (approved):

PR #129, opened 2023-02-28
docs: default value of SSL_ENGINE is `on`

theseion:
The default value of SSL_ENGINE was never correct in the readme.


fzipi (approved):

fzipi:
We might need to do the same in the `README-containers.md` now.

PR #132, opened 2023-03-09
fix: update apache because of CVE-2023-25690

fzipi:
Fixes CVE-2023-25690.


theseion (approved):
Using the full version already paid off :)

PR #134, opened 2023-03-18
chore: update activeate-rules.sh to match crs-setup.conf.example

theseion:


fzipi (approved):
LGTM. Is there a way we can fail if we detect a problem, so this doesn't happen again?

theseion:
Not really, because we target both pre and post v4, so there's always one that doesn't match. But at least the matching is not tolerant to changes in the formatting.

PR #141, opened 2023-04-16
Add all headers from example config

fzipi:
Update documentation with default values from CRS.


fzipi:
We should be using v3.3.4 values, and per https://github.com/coreruleset/coreruleset/blob/v3.3/master/crs-setup.conf.example#L482, we don't have these values there.

Closing.

PR #147, opened 2023-05-25
Add env variable `REAL_IP_PROXY`

ppp0:
Needed to be able to differentiate between `proxy_set_header` vs `real_ip` in nginx/templates/includes/proxy_backend.conf.template

Rationale:

Documented in https://nginx.org/en/docs/http/ngx_http_realip_module.html#example
```
set_real_ip_from  192.168.1.0/24;
set_real_ip_from  192.168.2.1;
set_real_ip_from  2001:0db8::/32;
real_ip_header    X-Forwarded-For;
real_ip_recursive on;
```

Still, `X-Real-IP` is to be passed to the proxy, as `$remote_addr` is going to be overwritten by the realip module
```
proxy_set_header X-Real-Ip $remote_addr;
```


theseion:
Thanks for the PR. I'm confused though: why does `REAL_IP_HEADER` not work for you? We made that configurable on purpose.

ppp0:
I needed something like this which was not possible with only one env var

```
set_real_ip_from  192.168.1.0/24;
set_real_ip_from  192.168.2.1;
set_real_ip_from  2001:0db8::/32;
real_ip_header    X-Forwarded-For;   # replaced from `$REAL_IP_HEADER`
real_ip_recursive on;

[..]

proxy_set_header X-Real-Ip $remote_addr;  #replaced from `$REAL_IP_PROXY`

```

theseion:
I see, thanks.
Do you think it makes more sense the way you've written it now than to use a simple boolean? AFAICT, this line would always be the same, unless you wouldn't want to set the header.

ppp0:
> [..] than to use a simple boolean? AFAICT, this line would always be the same, unless you wouldn't want to set the header.

I'm confused, sorry. Can you give me a code example or explain again?

theseion:
Your suggestion in pseudo code:
```
REAL_IP_PROXY="$remote_addr"
...
replace("{{REAL_IP_PROXY}}", "proxy_set_header X-Real-Ip {{REAL_IP_PROXY}};")
```

My suggestion in pseudo code:
```
REAL_IP_PROXY_ENABLED="on"
...
if (REAL_IP_PROXY_ENABLED == "on") {
  print("proxy_set_header X-Real-Ip $remote_addr;")
}
```

My question is, whether it makes sense to give users control over the value of `X-Real-Ip` in this case. I suspect that the value will always be `$remote_addr`.

ppp0:
> Your suggestion in pseudo code:
> 
> ```
> REAL_IP_PROXY="$remote_addr"
> ...
> replace("{{REAL_IP_PROXY}}", "proxy_set_header X-Real-Ip {{REAL_IP_PROXY}};")
> ```

Afraid, **I did non suggest that** but rather
```
 replace("{{REAL_IP_PROXY}}", "proxy_set_header {{REAL_IP_PROXY}} $remote_addr;")
```


> My question is, whether it makes sense to give users control over the value of `X-Real-Ip` in this case. I suspect that the value will always be `$remote_addr`.

With the current implementation, if you set `REAL_IP_HEADER`, both `real_ip_header` and `proxy_set_header` will be set. See line 5 and line 20 in https://github.com/coreruleset/modsecurity-crs-docker/blob/be5441fd00f78e2c2f32bdd845cb4daeec4c1b20/nginx/templates/includes/proxy_backend.conf.template

Real IP is a [nginx module](https://nginx.org/en/docs/http/ngx_http_realip_module.html) which overwrites `$remote_addr` ("client address" in nginx docu), where as `proxy_set_header` just instructs nginx to pass a header to the upstream node.

Honestly, for not adding complexity, I would stick to use env variables to only pass values, instead of a flag

May be `REAL_IP_PROXY`  is unfortunate in the naming, as it's an arbitrary name of a header which carries `$remote_addr`, but I haven't come up with anything better

README-containers.md
```diff
@@ -123,6 +123,7 @@ Note: Apache access and metric logs can be disabled by exporting the `nologging=
 | NGINX_ALWAYS_TLS_REDIRECT | if http should redirect to https (Allowed values: `on`, `off`. Default: `off`) |
 | SET_REAL_IP_FROM | A string of comma separated IP, CIDR, or UNIX domain socket addresses that are trusted to replace addresses in `REAL_IP_HEADER` (Default: `127.0.0.1`). See [set_real_ip_from](http://nginx.org/en/docs/http/ngx_http_realip_module.html#set_real_ip_from) |
 | REAL_IP_HEADER | Name of the header containing the real IP value(s) (Default: `X-REAL-IP`). See [real_ip_header](http://nginx.org/en/docs/http/ngx_http_realip_module.html#real_ip_header) |
+| REAL_IP_PROXY | Name of the header containing `$remote_addr` to be passed to proxy (Default: `X-REAL-IP`). See [proxy_set_header](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header) |
```
theseion, line 126:
```suggestion
| REAL_IP_PROXY_HEADER | Name of the header containing `$remote_addr` to be passed to proxy (Default: `X-REAL-IP`). See [proxy_set_header](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header) |
```

nginx/Dockerfile
```diff
@@ -140,6 +140,7 @@ ENV ACCESSLOG=/var/log/nginx/access.log \
     NGINX_ALWAYS_TLS_REDIRECT=off \
     SET_REAL_IP_FROM="127.0.0.1" \
     REAL_IP_HEADER="X-REAL-IP" \
+    REAL_IP_PROXY="X-REAL-IP" \
```
theseion, line 143:
```suggestion
    REAL_IP_PROXY_HEADER="X-REAL-IP" \
```

nginx/Dockerfile-alpine
```diff
@@ -134,6 +134,7 @@ ENV ACCESSLOG=/var/log/nginx/access.log \
     NGINX_ALWAYS_TLS_REDIRECT=off \
     SET_REAL_IP_FROM="127.0.0.1" \
     REAL_IP_HEADER="X-REAL-IP" \
+    REAL_IP_PROXY="X-REAL-IP" \
```
theseion, line 137:
```suggestion
    REAL_IP_PROXY_HEADER="X-REAL-IP" \
```

nginx/templates/includes/proxy_backend.conf.template
```diff
@@ -2,7 +2,7 @@ proxy_set_header Host $host;
 proxy_set_header Proxy "";
 proxy_set_header Upgrade $http_upgrade;
 proxy_set_header Connection $connection_upgrade;
-proxy_set_header ${REAL_IP_HEADER} $remote_addr;
+proxy_set_header ${REAL_IP_PROXY} $remote_addr;
```
theseion, line 5:
```suggestion
proxy_set_header ${REAL_IP_PROXY_HEADER} $remote_addr;
```

theseion (changes requested):

theseion:
Thanks for the clarification. Your changes make perfect sense.

nginx/Dockerfile
ppp0: Uh oh, build now fails with
```
README-containers.md contains 25041 characterswhich exceeds the 25000 character limit of Docker Hub
```
(https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/5174520958/jobs/9320929474?pr=147)

README-containers.md
ppp0: Uh oh, build now fails with
```
README-containers.md contains 25041 characters which exceeds the 25000 character limit of Docker Hub
```
(https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/5174520958/jobs/9320929474?pr=147)

We didn't add that many chars though :thinking:

theseion:
@fzipi I propose to remove the environment variables from the Docker Hub documentation. Do you agree?

ppp0:
> @fzipi I propose to remove the environment variables from the Docker Hub documentation. Do you agree?

@theseion I would not. This 25k limit on Docker Hub is a pity but, personally, I value seeing the configurables of an image while surfing around there. I'll try to shorten the text otherwise, if you agree

theseion:
Sure, give it a try. The reason I made the proposal is that I suspect we may add more variables in the future and will run into the same issue over and over.

ppp0:
@theseion Please review the change in wording https://github.com/coreruleset/modsecurity-crs-docker/pull/147/commits/9017a55519f0023ac5b366ece35a29e2b94d928d

README-containers.md
```diff
@@ -9,7 +9,7 @@
 
 ## Full documentation
 
-⚠️ We are limited to 25000 chars in the Docker Hub documentation. The full documentation is hosted on [GitHub](https://github.com/coreruleset/modsecurity-crs-docker/blob/master/README.md). There you can find how to build the containers for multiple architectures, how to add your own CRS version, and additional information.
+⚠️ We are limited to 25000 chars in the Docker Hub documentation. The full documentation is hosted on [GitHub](https://github.com/coreruleset/modsecurity-crs-docker/blob/master/README.md). There you can find additional in-depth information.
```
theseion, line 12:
```suggestion
⚠️ We are limited to 25000 chars in the Docker Hub documentation. The full documentation is hosted on [GitHub](https://github.com/coreruleset/modsecurity-crs-docker/blob/master/README.md).
```

theseion (changes requested):

theseion:
Thanks @ppp0, I've proposed a few adjustments. Would you mind also updating the same sentences in the full readme?

theseion:
Please rebase onto the latest changes. And would you mind updating the same sentences in the full README?

ppp0:
Ooops, of course, thanks

theseion (approved):

theseion:
Thanks!

PR #148, opened 2023-05-25
Update nginx to version 1.24.0

Boschman:


theseion (approved):
LGTM

fzipi (changes requested):
We always target `stable` versions. Are we going to use `latest` now? Stable is `1.24.0`.

theseion:
Oh, sorry. I hadn't realised that. @Boschman please change the version to `1.24.0`.

Boschman:
Good point!

@theseion, I've changed the version to 1.24.0.

fzipi (approved):

Issue #143, opened 2023-04-24
Modify "server" header


dontcrash:
It is possible to use SecServerSignature/ServerSignature to modify the server header to not give potential attackers extra info?

theseion:
I think I'd rather add three new environment variables `APACHE_SERVER_TOKENS`, `APACHE_SERVER_SIGNATURE`, and `MODSEC_SERVER_SIGNATURE` and give the user full control over the proxy.
And then, of course, an additional variable `UNSET_PROXIED_HEADERS` (or something like that), that takes a list of headers to unset, as you suggest. Ideally, this could also be used to achieve a similar effect for nginx.

franbuehler:
Good suggestion, thank you! 
To be honest, I was thinking of environment variables too. 
I'll try and open a PR for that. I would like to work with our containers a bit more again. I'll self assign this issue.

franbuehler:
I have a PR ready, but I'm not able to build the container locally to test it. 
I'll ask you or Max on Slack...

theseion:
I think this issue should have been fixed with #151. Not sure if something's still missing though.

fzipi:
Ping @franbuehler ...

audunru:
For nginx, I've been able to build additional modules using a Dockerfile such as this to get rid of the Server header. It's based on the modsecurity-crs-docker image, and builds the module using the same version of nginx so that you don't run into problems with different versions that you can get if you try to install the modules using apk.

This example is using [ngx_security_headers](https://github.com/GetPageSpeed/ngx_security_headers), but I've also tried headers-more and it works just as well.

```Dockerfile
ARG MODSECURITY_TAG="3.3-nginx-alpine-202312070812"

FROM owasp/modsecurity-crs:${MODSECURITY_TAG} as build

ARG SECURITY_HEADERS_VERSION="0.1.0"

RUN apk add --no-cache alpine-sdk pcre-dev pcre2-dev

RUN wget http://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz
RUN tar -xzvf nginx-${NGINX_VERSION}.tar.gz

RUN wget https://github.com/GetPageSpeed/ngx_security_headers/archive/refs/tags/${SECURITY_HEADERS_VERSION}.tar.gz
RUN tar -xzvf ${SECURITY_HEADERS_VERSION}.tar.gz

RUN cd nginx-${NGINX_VERSION}; \
  ./configure --with-compat \
[...]

PR #151, opened 2023-06-18
Add control to server signature and tokens

franbuehler:
This PR adds 3 new configurable env variables for Apache only:

`SERVER_TOKENS`: This directive was already set in apache/Dockerfile, but not configurable. It was always set to Prod. Default now is `Full` so that SERVER_SIGNATURE works.
`SERVER_SIGNATURE`: This new directive is configurable and allows to set server info in documents.
`MODSEC_SERVER_SIGNATURE`: When SERVER_TOKENS is set to full, the server response header can be changed.

I'm not able to build and test this locally but I'll do when I get some help.
I have to test if the server response header set by a backend can also be changed with `MODSEC_SERVER_SIGNATURE`. If not, I'll have to add a configurable `header unset`. Update: The server response header cannot be unset! It has to be set to Apache.


franbuehler:
The build error is the same I got locally.
What do I have to change? What's the problem?

https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/5302835017/jobs/9598050881?pr=151#step:6:275
and later:


`Error: buildx bake failed with: ERROR: failed to solve: process "/bin/sh -c set -eux;     echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections;     apt-get update -qq;     apt-get install -y -qq --no-install-recommends --no-install-suggests         automake         ca-certificates         g++         git         libapr1-dev         libaprutil1-dev         libcurl4-gnutls-dev         libfuzzy-dev         libpcre++-dev         libtool         libxml2-dev         libyajl-dev         lua5.2-dev         make         pkgconf         wget" did not complete successfully: exit code: 100`

theseion:
The new Debian version no longer has a package called `libpcre++-dev`. Trying to fix it now.

theseion:
Fix incoming: #152

franbuehler:
Thank you for the fix @theseion.
I'll try to test again...

fzipi:
Hi @franbuehler! What is the status on this one? Do you need help to setup a test environment?

franbuehler:
I'll work on this and test it after my vacation.
Thanks for the ping.

franbuehler:
I was finally able to build and test the image. The directive `SecServerSignature` is not meant to work in reverse proxy mode. But we could work with `Header set Server: Apache` to achieve the goal to override the Server response header set by the backend. I will adjust the PR.

theseion:
I think it makes more sense to support setting arbitrary headers. Of course, that needs to be encoded somehow. I suggest commas to separate entries and colons to separate header names from values, as that is what's used in HTTP anyway. That means, you would add a new environment variable that could be configured like so:

```
SET_HEADERS="Accept:*/*,Content-Type:application/json"
```

You would also need to add a shell script to parse the variable and transform its value into httpd / nginx expressions to set headers. While you're at it, it would make sense to do the same thing to unset headers.
Actually, to solve the original issue, we probably only need a way to unset headers, right? That would be a bit simpler.

Both images have a mechanism to apply environment variables to configurations. As long as you specify the environment variable in the configuration and the set it appropriately, that part should work. The real work is transforming the user input into the correct form.

theseion:
For nginx you'd create a script like `92-update-real-ip.sh`.
For httpd we don't yet have custom scripts. I'd add the same structure as for nginx and then add the script there, so `docker-entrypoint.d/yourscript.sh`; then call the script from `docker-entrypoint.sh`.

fzipi:
@franbuehler Any news on this one?

franbuehler:
This is getting a bit complex now.
My suggestion is:
I'll finish this PR that will provide 3 new, configurable directives to Apache only:

- `MODSEC_SERVER_SIGNATURE`  -> `SecServerSignature` in apache/conf/extra/httpd-modsecurity.conf / [not supported in v3](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v3.x%29#secserversignature) -> I will reflect that in the README.
- `SERVER_SIGNATURE` -> `ServerSignature` in Apache. Not available in NGINX. I will rename the var to APACHE_SERVER_SIGNATURE and reflect that in the README.
- `SERVER_TOKENS` -> `ServerTokens` in Apache -> `server_tokens off;` in NGINX does not have the same options and has a different meaning -> therefore I'll open a new PR for NGINX. I will rename the var to APACHE_SERVER_TOKENS and describe that in the README.

This means, we can review and merge this PR (as soon as I fixed it) with the knowledge that it does not solve issue #143, but provides 3 new directives to Apache.

So my plan is:
1. to finish and get this PR merged
2. open a new PR that adds `server_tokens off;` (configurable) to NGINX
3. open a PR that provides the functionalities (set/unset headers) suggested by @theseion

theseion (approved):
LGTM. Thanks!

PR #157, opened 2023-08-07
feat: add variable to configure httpd mutexes

theseion:
Add `MUTEX` variable to enable configuration of httpd mutexes.


theseion:
Context: httpd on Apple silicon has a bug. It will only start when the mutex is set to `posixsem`.

fzipi (approved):

PR #169, opened 2023-10-07
feat: add server_tokens to nginx images

franbuehler:
This PR adds the nginx directive `server_tokens` to our nginx images and solves part 2 of issue #143.

Full plan:
https://github.com/coreruleset/modsecurity-crs-docker/pull/151#issuecomment-1705366720

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ docker run -p 8080:80 -e SERVER_NAME=myhost my-modsec
 | SERVER_ADMIN  | A string value indicating the address where problems with the server should be e-mailed (Default: `root@localhost`) |
 | SERVER_NAME | A string value indicating the server name (Default: `localhost`) |
 | SERVER_SIGNATURE | A string value configuring the footer on server-generated documents (Default: `Off`) |
-| SERVER_TOKENS | A string value configuring the Server HTTP response header. Also see MODSEC_SERVER_SIGNATURE. (Default: `Full`). |
+| SERVER_TOKENS | A string value indicating the server header of HTTP responses. Also see MODSEC_SERVER_SIGNATURE. (Default: `Full`). |
 | SSL_CIPHER_SUITE | A string indicating the cipher suite to use. Uses OpenSSL [list of cipher suites](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_ciphersuites.html) (Default: `"ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384"` |
 | SSL_ENGINE  | A string indicating the SSL Engine Operation Switch (Default: `on`) |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -171,7 +171,7 @@ docker run -p 8080:80 -e SERVER_NAME=myhost my-modsec
 | SERVER_ADMIN  | A string value indicating the address where problems with the server should be e-mailed (Default: `root@localhost`) |
 | SERVER_NAME | A string value indicating the server name (Default: `localhost`) |
 | SERVER_SIGNATURE | A string value configuring the footer on server-generated documents (Default: `Off`) |
-| SERVER_TOKENS | A string value configuring the Server HTTP response header. Also see MODSEC_SERVER_SIGNATURE. (Default: `Full`). |
+| SERVER_TOKENS | A string value indicating the server header of HTTP responses. Also see MODSEC_SERVER_SIGNATURE. (Default: `Full`). |
```
theseion, line 174:
```suggestion
| SERVER_TOKENS | Option defining the server information presented to clients in the `Server` HTTP response header. Also see `MODSEC_SERVER_SIGNATURE`. (Allowed values: `Full`,  `Prod[uctOnly]`, `Major`, `Minor`, `Min[imal]`, `OS`. Default: `Full`). |
```

README.md
```diff
@@ -171,7 +171,7 @@ docker run -p 8080:80 -e SERVER_NAME=myhost my-modsec
 | SERVER_ADMIN  | A string value indicating the address where problems with the server should be e-mailed (Default: `root@localhost`) |
 | SERVER_NAME | A string value indicating the server name (Default: `localhost`) |
 | SERVER_SIGNATURE | A string value configuring the footer on server-generated documents (Default: `Off`) |
```
theseion, line 173:
```suggestion
| SERVER_SIGNATURE | A string value configuring the footer on server-generated documents (Allowed values: `On`, `Off`, `EMail`. Default: `Off`) |
```

README.md
```diff
@@ -213,6 +213,7 @@ Note: Apache access and metric logs can be disabled by exporting the `nologging=
 | PROXY_SSL_PROTOCOLS | A string value indicating the ssl protocols to enable (default: `TTLSv1.2 TLSv1.3`)|
 | PROXY_SSL_VERIFY  | A string value indicating if the client certificates should be verified (Allowed values: `on`, `off`. Default: `off`) |
 | PROXY_TIMEOUT  | Number of seconds for proxied requests to time out connections (Default: `60s`) |
+| SERVER_TOKENS | A string value indicating the server header of HTTP responses and the version number in the footer on error pages. (Allowed values: `on`, `off`. Default: `off`). |
```
theseion, line 216: Setting an arbitrary string is only possible with a commercial nginx license.
```suggestion
| SERVER_TOKENS | A boolean value for enabling / disabling emission of server identifying information in the `Server` HTTP response header and on error pages. (Allowed values: `on`, `off`, `build`. Default: `off`). |
```

theseion (changes requested):

nginx/templates/conf.d/server_tokens.conf.template
theseion, line 1: Why did you opt to create a new file instead of adding the directive to `default.conf.template`?

theseion (changes requested):

nginx/templates/conf.d/server_tokens.conf.template
franbuehler: I've seen this option in one of the nginx images and I liked this modular approach. But I can change it.

theseion: Yes, it's nice, we just didn't write any of the other config files in a modular way, so there's no way to include that file in different contexts (`http`, `server`, etc.). I would argue that keeping the variables together currently has more value for w.r.t. our maintenance efforts.

franbuehler: Ok, I see, makes sense. Thank you for your explanation. I'll change that today or tomorrow.

theseion (approved):

Issue #182, opened 2024-01-12
[FEATURE] add vi tool


leveryd:
```
➜  ~ docker run -ti --rm owasp/modsecurity-crs:3.3.5-nginx-alpine-202401080101
/etc/modsecurity.d # vi --help
BusyBox v1.36.1 (2023-11-06 11:32:24 UTC) multi-call binary.

Usage: vi [-c CMD] [-R] [-H] [FILE]...

Edit FILE

	-c CMD	Initial command to run ($EXINIT and ~/.exrc also available)
	-R	Read-only
	-H	List available features
```

vi is "busybox vi", it can only support basic editor and is not enough. 

For example, "busybox vi" do not support "V" mode.

fzipi:
IMHO, if you need an editor in a production image, something is broken.

Do you have a strong requirement on this, or better to install it when needed for your use case (instead of forcing the whole community to have it)?

leveryd:
Yes,i have strong requirement on this.

I use this docker image to test some rules, so i need ofen to modify the rule files. I know there are some ways to edit rule file instead of using `busybox vi`, but it is better for me if i can edit it directly in the container.

fzipi:
No worries. As this image is being used by folks in production, I would rather not add full vim/vi dependencies as not many people will use them.

Buy you are free to create your own image based on the ones we provide and add your preferred editor. It is just creating a new layer using docker's own `FROM`...

PR #183, opened 2024-01-12
Update Dockerfile-alpine

leveryd:
https://github.com/coreruleset/modsecurity-crs-docker/issues/182


theseion:
The associated issue was labeled as wontfix.

PR #187, opened 2024-01-19
Updated docker buildx installation link

Xhoenix:


README.md
```diff
@@ -41,7 +41,7 @@ docker buildx create --use --platform linux/amd64,linux/i386,linux/arm64,linux/a
 docker buildx bake -f docker-bake.hcl
 ```
 
-We require a version of `buildx` >= v0.9.1. [Visit the official documentation](https://docs.docker.com/build/buildx/install/) for instructions on installing and upgrading `buildx`. You can check which version you have using:
+We require a version of `buildx` >= v0.9.1. [Visit the official documentation](https://github.com/docker/buildx#installing) for instructions on installing and upgrading `buildx`. You can check which version you have using:
```
theseion, line 44:
```suggestion
We require a version of `buildx` >= v0.9.1. [Visit the official documentation](https://docs.docker.com/build/architecture/#install-buildx) for instructions on installing and upgrading `buildx`. You can check which version you have using:
```

theseion (approved):

README.md
Xhoenix: The buildx GH repo contains *direct* install instructions. Is there any reason Docker docs given preference?

theseion: You're right. The reason I opted for the documentation page is that it says there that you don't actually need to install anything with newer Docker versions. The GitHub page doesn't do that, so people that don't know what buildx is might try to install it even though they wouldn't need to.

PR #188, opened 2024-01-24
feat: drop support for overriding USER

### LINKED ISSUE #184: USER does not work.
```
docker run -e USER=root -ti --rm owasp/modsecurity-crs:3.3.5-nginx-alpine-202401080101

➜  ~ docker exec -ti 2d9c96809969 sh
/usr/share/nginx/html # ps aux
PID   USER     TIME  COMMAND
    1 root      0:00 nginx: master process nginx -g daemon off;
  125 nginx     0:00 nginx: worker process
  126 nginx     0:00 nginx: worker process
  127 nginx     0:00 nginx: worker process
  128 nginx     0:00 nginx: worker process
  129 root      0:00 sh
  136 root      0:00 ps aux
```

I have set USER=root , but the nginx process user is still nginx.

theseion:
Overrideing the USER only makes sense if it can be configured at build time, as that user needs to exist on the system. For containers, this functionality doesn't make much sense anyway.

Fixes #184

### DIFF
--- a/README-containers.md
+++ b/README-containers.md
@@ -174,8 +174,6 @@ All these variables impact in configuration directives in the modsecurity engine
 
 | Name | Description|
 | -- | -- |
-| USER | Name (or #number) of the user to run httpd or nginx as (Default: `www-data` (httpd), `nginx` (nginx)) |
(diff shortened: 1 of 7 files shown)


fzipi (approved):
Excellent.

PR #189, opened 2024-01-27
feat: move version information to bake file

theseion:


fzipi (approved):
Thanks, this clearly will simplify builds!

PR #191, opened 2024-01-28
feat: use nginx-unprivileged images

### LINKED ISSUE #55: Running with non-root access with the nginx user
If anyone else is wondering how to use this image for in a non-root context then what worked for me in the Dockerfile:
```
...

# chown all modsecurity and nginx config files to the nginx user
RUN chown -R nginx:nginx /etc/nginx/
RUN chown -R nginx:nginx /opt/modsecurity/
RUN chown -R nginx:nginx /etc/modsecurity.d/
RUN chown -R nginx:nginx /opt/owasp-crs/
RUN chown -R nginx:nginx /usr/local/


USER nginx
...
```

theseion:
Use unprivileged images and always run images as non-root.

Fixes #55

### DIFF
--- a/apache/Dockerfile
+++ b/apache/Dockerfile
@@ -58,8 +58,8 @@ RUN set -eux; \
 
 FROM httpd:${HTTPD_VERSION}
 
-ARG MODSEC2_VERSION="n/a"
-ARG LUA_VERSION="n/a"
+ARG MODSEC2_VERSION
+ARG LUA_VERSION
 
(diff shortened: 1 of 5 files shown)


fzipi (approved):
AWESOME

fzipi:
QQ: does this look like a breaking change? 🤔

theseion:
Not necessarily. There will probably be some for people using mounts / overrides, or for those using one of our images as a basis.

theseion:
I just ran the CRS test suite using these images and didn't spot any change in behaviour.

PR #193, opened 2024-01-29
feat: publish images to ghcr.io as well

theseion:


fzipi:
Hmmm... doesn't publish need to have a dependency on verify? From what I see here, images are going to be published just when it builds ok, but it doesn't mean they are working...

I would at least publish if images are verified...

theseion:
We could do that. The way it's set up at the moment though, the verification workflow would always run before the publishing workflow, because we merge from `develop` to `master`, and `develop` has already been verified because we merge PRs to `develop`.
On the other hand, I suppose it doesn't hurt to verify images before publishing one more time. I would, however, change the way it's done at the moment. Right now, as you say, verification runs in parallel with publishing. Even if we serialize those jobs, both workflows would currently run full builds because they don't run within the same context. I could try to refactor that into three jobs:
- build: build all images and create tarball outputs for the following jobs
- verify: verify images from the build job
- publish: publish images from the build job

That looks beautiful but seems like a lot of work to me. My two cents: because we never push to `master`, I think we can simply refrain from running verification on the `master` branch (again).

fzipi:
Ok, makes sense. So do you want to merge this one?

PR #200, opened 2024-02-12
feat: add cosign

### LINKED ISSUE #96: Sign images using cosign
Check what is needed to sign images using cosign.

References
- https://github.com/sigstore/cosign-installer

fzipi:
## what

- add cosign to sign images
- per https://docs.docker.com/build/ci/github-actions/push-multi-registries/, just using the proper tags should push to both registries so we don't need to build twice

Fixes #96

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -31,12 +31,12 @@ variable "lmdb-version" {
     default = "0.9.29"
 }
 
-variable "REPO" {
-    default = "owasp/modsecurity-crs"
-}
-
-variable "GHCR_REPO" {
-    default = "ghcr.io/coreruleset/modsecurity-crs"
+variable "REPOS" {
+    # List of repositories to tag
+    default = [
+        "owasp/modsecurity-crs",
+        "ghcr.io/coreruleset/modsecurity-crs",
+    ]
 }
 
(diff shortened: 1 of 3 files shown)


fzipi:
See https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/7873757487/job/21481826636?pr=200 for metadata output used for cosign step.

docker-bake.hcl
```diff
@@ -52,7 +56,7 @@ function "patch" {
 
 function "tag" {
     params = [tag]
-    result = ["${REPO}:${tag}"]
+    result = ["${REPO}:${tag}", "${GHCR_REPO}:${tag}"]
```
theseion, line 59: I think this defeats the purpose of the the `REPO` variable. If a user that is not use want's to tag a build with their own repository, specifying `REPO` will still also tag GH. Maybe make the two repos fixed and default and allow users to supply a list of repositories instead?

theseion (changes requested):

docker-bake.hcl
fzipi: Sounds good.

theseion (approved):

Issue #201, opened 2024-02-13
Latest nginx build not working


guidugli:
I was reinstalling a server and noticed that the latest image is not working. Doing some tests I noticed that running the command below on a 3 month old image works, but it fails with the latest one.

`podman run --name abc -p 8080:8080 -p 8443:443 -ti -e PROXY=0 -e ALLOWED_METHODS="GET POST PUT"  -e MODSEC_RULE_ENGINE=Off  --rm docker.io/owasp/modsecurity-crs:nginx`

The error happens on both arm and x64.

Here is the log when running the latest image:

> /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
> /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
> /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-generate-certificate.sh
> /usr/local/bin/generate-certificate: generating new certificate
> Warning: No -copy_extensions given; ignoring any extensions in the request
> /usr/local/bin/generate-certificate: generated /etc/nginx/conf/server.key and /etc/nginx/conf/server.crt
> /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
> 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
> 10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
> /docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
> /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
> 20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/modsecurity.d/modsecurity-override.conf.template to /etc/nginx/modsecurity.d/modsecurity-override.conf
> 20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/modsecurity.d/setup.conf.template to /etc/nginx/modsecurity.d/setup.conf
> 20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/conf.d/default.conf.template to /etc/nginx/conf.d/default.conf
> 20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/conf.d/logging.conf.template to /etc/nginx/conf.d/logging.conf
[...]

PR #202, opened 2024-02-13
fix: use higher port for nginx running unprivileged

fzipi:
Might solve #201.


README.md
```diff
@@ -145,6 +145,8 @@ An example can be seen in the [docker-compose](https://github.com/coreruleset/mo
 > 💬 What happens if I want to make changes in a different file, like `/etc/nginx/conf.d/default.conf`?
 > You mount your local file, e.g. `nginx/default.conf` as the new template: `/etc/nginx/templates/conf.d/default.conf.template`. You can do this similarly with other files. Files in the templates directory will be copied and subdirectories will be preserved.
 
+Nginx also run as **unprivileged user**. This means that we cannot bind to ports below 1024, so you might need to correct your PORT and SSL_PORT settings. Now the defaults are 8080 and 8443.
```
theseion, line 148:
```suggestion
nginx is run with an **unprivileged user**. This means that we cannot bind to ports below 1024, so you might need to correct your `PORT` and `SSL_PORT` settings. Now the defaults for nginx are `8080` and `8443`.
```

theseion (changes requested):

theseion (approved):

PR #207, opened 2024-02-15
fix: OpenResty image not running

matias-at-skribble:
- fix paths on Dockerfile (fix `nginx: [emerg] cannot load certificate "/etc/nginx/conf/server.crt"`)
- add missing TLS config script (fix `nginx: [emerg] unknown directive "PROXY_SSL_CONFIG" in /usr/local/openresty/nginx/conf/conf.d/default.conf:19`)


fzipi (approved):
Thanks @matias-at-skribble for your contribution!

LGTM.

PR #208, opened 2024-02-15
fix: add lua modules to openresty

fzipi:


theseion:
Oh, was this your question earlier, whether to use `lua-rocks` for OpenResty to install the modules? For the OpenResty image that may be the better option, TBH. I'd prefer all packages to come from the same source (in this case, all lua modules from the Lua package manager). But it probably doesn't make huge difference.

fzipi:
Yes, exactly. That was my question. No worries, I'll change this one.

openresty/Dockerfile-alpine
```diff
@@ -113,6 +113,7 @@ RUN set -eux; \
 FROM openresty/openresty:${OPENRESTY_VERSION}-alpine-fat
 
 ARG MODSEC3_VERSION
+ARG LUA_ROCKS_MODULES
```
theseion, line 116:
```suggestion
ARG LUA_MODULES
```

theseion (changes requested):

docker-bake.hcl
```diff
@@ -27,6 +27,13 @@ variable "lua-version" {
     default = "5.3"
 }
 
+variable "lua-modules" {
+  default = [
+    "lua-lzlib",
```
theseion, line 32: Wrong package name for Debian. We need to separate Alpine and Debian modules.

openresty/Dockerfile-alpine
```diff
@@ -232,7 +233,9 @@ RUN set -eux; \
         sed \
         yajl; \
     # Install lua modules here
-    luarocks install lua-resty-openidc; \
+    for module in lua-resty-openidc ${LUA_MODULES}; do \
```
theseion, line 236: Now I think it makes sense to move `lua-resty-openidc` to the bake file as well.

fzipi:
Well, I would love to see the maintainer @TafkaMax chiming on this in the future. We mentioned that we wouldn't maintain the image, and I think this was more work than expected for openresty.
- git must be installed for luarocks to build `lua-zlib`. This pointed me to a problem in our alpine image, where the module `lua-lzlib` is different from `lua-zlib`, and it is unmaintained and the repo is archived.
- then everything is built in the last layer. I think the alpine-fat might not make sense if there is one additional step building all the lua modules and then copying the results to the final image. My 2 cents.

👉 In the end, if there is no maintenance for this image, I would drop it. We don't need the additional work. 🤷

theseion (approved):

theseion:
Agreed. Let's give them the benefit of the doubt for now ;)

Issue #172, opened 2023-11-22
Not running with readOnlyRootFilesystem set to true


enibache:
I'm trying to run this image in my helm chart running in OCP4 whereby we have a quality gate that only allows pods with readOnlyRootFilesystem set to true. Is there perhaps hints on how to accomplish this using this image in my deployment.yaml?

enibache:
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-generate-certificate.sh
/usr/local/bin/generate-certificate: using existing key / certificate
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/conf.d/default.conf.template to /etc/nginx/conf.d/default.conf
/docker-entrypoint.d/20-envsubst-on-templates.sh: line 33: can't create /etc/nginx/conf.d/default.conf: Read-only file system

fzipi:
There are two problems: 
- how we create the container image, which for the nginx case needs to write files around.
- the logs must be written somewhere

The second one can be solved by using a persistent volume like:

fzipi:
The [nginx ingress controller](https://docs.nginx.com/nginx-ingress-controller/configuration/security/) solves the first part by ... copying everything to a mounted directory:
```
#      volumes:
#      - name: nginx-etc
#        emptyDir: {}
#      - name: nginx-cache
#        emptyDir: {}
#      - name: nginx-lib
#        emptyDir: {}
#      - name: nginx-log
#        emptyDir: {}
.
.
.
#          readOnlyRootFilesystem: true
.
.
.
#        volumeMounts:
#        - mountPath: /etc/nginx
#          name: nginx-etc
#        - mountPath: /var/cache/nginx
#          name: nginx-cache
#        - mountPath: /var/lib/nginx
#          name: nginx-lib
#        - mountPath: /var/log/nginx
#          name: nginx-log
.
.
.
#      initContainers:
[...]

PR #210, opened 2024-02-18
feat: enable running nginx containers with read-only root fs

theseion:
Refs #172

### DIFF
--- a/apache/Dockerfile
+++ b/apache/Dockerfile
@@ -1,6 +1,6 @@
 ARG HTTPD_VERSION="n/a"
 
-FROM httpd:${HTTPD_VERSION} AS build
+FROM image AS build
 
 ARG MODSEC2_VERSION="n/a"
 ARG LUA_VERSION="n/a"
@@ -36,7 +36,7 @@ RUN set -eux; \
     make install; \
     make clean
 
-FROM httpd:${HTTPD_VERSION} AS crs_release
+FROM image AS crs_release
 
 ARG CRS_RELEASE
 
@@ -56,7 +56,7 @@ RUN set -eux; \
     rm -f v${CRS_RELEASE}-minimal.tar.gz coreruleset-${CRS_RELEASE}-minimal.tar.gz.asc; \
     mv -v /opt/owasp-crs/crs-setup.conf.example /opt/owasp-crs/crs-setup.conf
 
-FROM httpd:${HTTPD_VERSION}
+FROM image
 
 ARG MODSEC2_VERSION
 ARG LUA_VERSION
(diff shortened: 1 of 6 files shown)


theseion:
@ne20002 @enibache This PR enables the use of read-only root fs. It's certainly not perfect but I din't want to rewrite the entire build setup and documentation. So for now you can run this image as follows with Docker:
```
docker run \
  --rm \
  --readonly \
  --tmpfs /etc/nginx:rw,mode=777 \
  --tmpfs /etc/modsecurity.d:rw,mode=777 \
  --tmpfs /opt/owasp-crs:rw,mode=777 \
  --tmpfs /tmp:rw,mode=777 \
  --tmpfs /var/cache/nginx:rw,mode=777 \
  owasp/modsecurity-crs:nginx
```bash
The container will copy files at startup to those `tmpfs` directories. In writeble root fs mode, the container behaves the same as before, in read-only mode, the `tmpfs` are required and the files will be copied there.

I'd appreciate some feedback.

You can build the image from this PR with (for example):
```bash
docker buildx bake --load --set "*.platform=linux/amd64" nginx
```

fzipi (commented):
Probably adding some documentation to the readme file would make sense for this case?

ne20002:
Sorry. I'm still fiddling with the 3.3.5 and can't get my exclusion for local ips get to work. Also, it seems as if there is no version 4 docker image available at hub.docker.com. I'd need to get it working first before I try to use it with a readonly setup.

theseion:
Thanks for the hint @ne20002. We thought we had published the images but really didn't. I've created the PR to fix that.

theseion:
@ne20002 The 4.0 images are now available.

theseion:
Well, I don't want to merge this without proper testing. Unless we create a root-only image variant, so that the other images continue to work. Maybe that would be the better approach anyway?

ne20002:
Hi @theseion 
I'm currently a bit busy. I got my Friendica setup working so far (still a few things to opimize) and will continue to update my Nextcloud setup (which seems to have a number of changes with the new plugin mechanism).

theseion:
@fzipi I think we should create a new image variant. Existing users may not want to or be able to switch anyway. Then it wouldn't be a tragedy if we still had some bugs in the new variant. What do you think?

fzipi:
Sounds a nice way to move forward.

theseion:
ping @fzipi

fzipi (approved):
This is getting complex. Nice solution, btw.

fzipi:
Are we following up with a docs update PR?

PR #216, opened 2024-03-08
fix: rearranged ENV variables in strict 'sort' order

dune73:
It started with move-mutex and then it evolved a bit (hence the branch name).


fzipi:
Are tables sorted in the readme*? Maybe we want to double check and sort everything...

dune73:
Good thinking. Done that for both READMEs.

fzipi (approved):
Glad that this was sorted out! 😛

PR #218, opened 2024-03-17
feat(apache): switch to pcre2

fzipi:
Use pcre2 for building on apache

### DIFF
--- a/apache/Dockerfile-alpine
+++ b/apache/Dockerfile-alpine
@@ -33,7 +33,7 @@ RUN set -eux; \
     make \
     openssl \
     openssl-dev \
-    pcre-dev \
+    pcre2-dev \
     zlib-dev
 
 RUN set -eux; \
@@ -41,7 +41,7 @@ RUN set -eux; \
     tar -zxvf v${MODSEC2_VERSION}.tar.gz; \
     cd ModSecurity-${MODSEC2_VERSION}; \
     ./autogen.sh; \
-    ./configure --with-yajl --with-ssdeep --with-lmdb; \
+    ./configure --with-yajl --with-ssdeep --with-lmdb --with-pcre2; \
     make; \
     make install; \
     make clean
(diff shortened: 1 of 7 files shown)


theseion:
Note that the failing test in the httpd builds is ok. The test has an override to make it work on the CI, which will no longer be necessary once we update to the new images. I'll push a change to CRS to update the override so this PR can be properly verified.

apache/Dockerfile-alpine
```diff
@@ -33,15 +33,15 @@ RUN set -eux; \
...
+    pcre2-dev \
     zlib-dev
 
 RUN set -eux; \
     wget --quiet https://github.com/owasp-modsecurity/ModSecurity/archive/refs/tags/v${MODSEC2_VERSION}.tar.gz; \
     tar -zxvf v${MODSEC2_VERSION}.tar.gz; \
     cd ModSecurity-${MODSEC2_VERSION}; \
     ./autogen.sh; \
-    ./configure --with-yajl --with-ssdeep --with-lmdb; \
+    ./configure --with-yajl --with-ssdeep --with-lmdb --with-pcre2; \
```
airween, line 44:
```suggestion
    ./configure --with-yajl --with-ssdeep --with-pcre2; \
```
mod_security2 module does not have `--with-lmdb` option.

airween (changes requested):

apache/Dockerfile-alpine
theseion: That flag is currently only used in the Alpine builds. Is that a mistake? Shouldn't both nginx builds have the flag?

theseion:
@fzipi ready for review.

airween (approved):
LGTM

fzipi:
We need an extra approver...

theseion:
I've added @airween to the containers team.

PR #220, opened 2024-03-26
Fixing part 7.12 of CIS Apache Benchmark - removing non-forward secrecy ciphers 

azurit:
Removing these non-forward secrecy ciphers from Apache configuration:
```
ECDHE-ECDSA-AES128-GCM-SHA256
ECDHE-ECDSA-AES256-GCM-SHA384
```


dune73:
Link to CRS wiki with the plan for this:

* https://github.com/coreruleset/coreruleset/wiki/Hardening%3A-the-Apache-Alpine-Docker-Container-2024#user-content-712-l2-ensure-only-cipher-suites-that-provide-forward-secrecy-are-enabled-automated

azurit:
Hmm, looks like i picked wrong cyphers - both of these supports forward secrecy. But i can't see those mentioned in CIS benchmark anywhere in the configuration.

fzipi:
@azurit @dune73 What's next here? Can you provide an update PR with the updated ciphers?

theseion (dismissed):

fzipi:
Checked this against the current defaults, and I think the conclusion @azurit already reached in March 2024 still holds: the two ciphers being removed do provide forward secrecy, so CIS 7.12 is met without this change.

Key exchange for every cipher in the current list:

```
ECDHE-ECDSA-AES128-GCM-SHA256    Kx=ECDH  Au=ECDSA
ECDHE-RSA-AES128-GCM-SHA256      Kx=ECDH  Au=RSA
ECDHE-ECDSA-AES256-GCM-SHA384    Kx=ECDH  Au=ECDSA
ECDHE-RSA-AES256-GCM-SHA384      Kx=ECDH  Au=RSA
ECDHE-ECDSA-CHACHA20-POLY1305    Kx=ECDH  Au=ECDSA
ECDHE-RSA-CHACHA20-POLY1305      Kx=ECDH  Au=RSA
DHE-RSA-AES128-GCM-SHA256        Kx=DH    Au=RSA
DHE-RSA-AES256-GCM-SHA384        Kx=DH    Au=RSA
```

Every one is ephemeral — `ECDHE`/`DHE` — so all eight are forward-secret. The `ECDSA` in the two removed suites is the *authentication* algorithm, not the key exchange, which is what the CIS control is about.

Removing them would also cost something: they are the only TLS 1.2 suites in the list that work with an ECDSA certificate. Anyone mounting an EC certificate through `SSL_CERT_FILE` would be left with no usable TLS 1.2 cipher.

The branch is also from before the SSL variable rename — it edits `SSL_CERT` and `PROXY_SSL_CERT`, which are now `SSL_CERT_FILE` and `PROXY_SSL_CERT_FILE` — hence the conflicts.

If a CIS gap remains in this area it would be worth a fresh issue with the specific finding, since the hardening plan in the [wiki page @dune73 linked](https://github.com/coreruleset/coreruleset/wiki/Hardening%3A-the-Apache-Alpine-Docker-Container-2024) covers more than ciphers. Thanks for raising it either way, @azurit — the check was worth running even though the list turned out to be clean.

fzipi:
Ran the live check as well, and it turns the earlier "this would cost something" into "this would break the default images".

**The certificate these images generate is ECDSA**, so on TLS 1.2 the only usable suites are the `ECDHE-ECDSA-*` ones this PR removes:

```
$ openssl s_client -connect <nginx>:8443 | openssl x509 -noout -text
    Public Key Algorithm: id-ecPublicKey
    Public-Key: (256 bit)

TLS 1.2, default certificate:
  ECDHE-ECDSA-AES128-GCM-SHA256    offered
  ECDHE-ECDSA-AES256-GCM-SHA384    offered
  ECDHE-ECDSA-CHACHA20-POLY1305    offered
  ECDHE-RSA-AES128-GCM-SHA256      refused   <- cannot be used with an EC certificate
```

Same on apache. With these two suites gone, a stock container has no TLS 1.2 cipher left for its own certificate; only TLS 1.3 clients would connect.

**And there is nothing non-forward-secret to remove.** To test that properly I mounted an RSA certificate, so that RSA key exchange would at least be possible, then probed:

```
TLS 1.2, RSA certificate:
  ECDHE-RSA-AES128-GCM-SHA256    [FS]      offered
  DHE-RSA-AES128-GCM-SHA256      [FS]      offered
  AES128-GCM-SHA256              [non-FS]  refused
  AES256-SHA                     [non-FS]  refused
  DES-CBC3-SHA                   [non-FS]  refused
  ECDH-RSA-AES128-GCM-SHA256     [non-FS]  refused
  RC4-SHA                        [non-FS]  refused
```

Nothing without forward secrecy negotiates, on either server, with either certificate type. CIS 7.12 passes as it stands.

PR #225, opened 2024-04-09
docs: document use of volume mounts for logs

theseion:


dune73:
Useful example. Thanks.

theseion:
Why don't we just make the Docker Hub readme very concise and link to the "full docs"? That way, people need to go to the GH readme.

theseion:
I've update the readme for Docker Hub. I noticed that the version on Docker Hub is outdated anyway. We don't push the readme automatically, it seems, so someone need to do it manually or we modify the job. As I no longer expect any changes to be made to the readme, we could stick to the manual approach, if we have the credentials.

fzipi (approved):

fzipi:
Manually pushed update.

PR #235, opened 2024-04-24
chore: error if port or ssl_port are lower than 1024

fzipi:
## what

- do not start container if port or ssl_port are defined lower than 1024

## why

after moving to unprivileged containers, people might not have updated their setups, bringing problems when starting


theseion (changes requested):

fzipi:
Maybe moving this to a file instead so it can be included in both servers.

theseion (approved):

PR #237, opened 2024-04-26
feat: manage deps with renovate

theseion:
Experimental support for updating versions in bake file.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,99 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "local>coreruleset/renovate-config",
+    "schedule:weekly"
+  ],
+  "enabledManagers": [
+    "custom.regex"
+  ],
+  "packageRules": [
+    {
+      "groupName": "all non-major dependencies",
+      "groupSlug": "all-minor-patch",
+      "matchPackagePatterns": [
+        "*"
+      ],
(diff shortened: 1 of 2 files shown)


fzipi (commented):
Looking good. Can you also include the common config using [this syntax](https://github.com/coreruleset/website/blob/main/renovate.json)?

The idea is to centralize renovate config in the https://github.com/coreruleset/renovate-config repo.

renovate.json
```diff
@@ -0,0 +1,126 @@
...
+      "depNameTemplate": "ModSecurity 3",
+      "customType": "regex",
+      "fileMatch": [
+        "^docker-bake.hcl$"
+      ],
+      "matchStrings": [
+        "^variable \"modsec3-version\" {\\n\\s+default = \"(?<currentValue> [^\"]+)\""
+      ],
+      "packageNameTemplate": "owasp-modsecurity/ModSecurity",
+      "datasourceTemplate": "github-releases"
```
fzipi, line 58: Is it smart enough to use the proper version? E.g. upgrading the corresponding major?

theseion: Yes, with the additional changes I've made.

theseion:
I've figured out how to restrict ModSecurity 2 and how to also update the README automatically. I've tested it locally, but we'll only see what PRs are actually created once it runs for the first time.

fzipi (approved):
Excellent, let's try this out! 🚀

PR #239, opened 2024-04-26
feat: override full modsecurity.conf file

### LINKED ISSUE #234: Review including modsecurity.conf-recommended or just use the override
In our apache build we are using the upstream `modsecurity.conf-recommended` file, and we add our own override file after that one.

This causes some friction because it always require files that are not going to be used, and also extended permissions when running in no permission filesystems.

For example:
```
apache-nightly                        | AH00526: Syntax error on line 199 of /etc/modsecurity.d/modsecurity.conf:
apache-nightly                        | ModSecurity: Failed to open the audit log file: /var/log/modsec_audit.log
```

This doesn't make sense because it never gets to our override that will actually point that file to the correct directory.

My proposal is to just replace it with the actual override instead, merging both files in our own file.

fzipi:
- move all variables from modsecurity-override.conf to modsecurity.conf
- add extra variables to configure modsecurity.conf params
- update README with new variables
- use only curl and remove wget usage (consistency)

Fixes #234


src/etc/modsecurity.d/modsecurity.conf
fzipi, line 1: This is based on latest v3 recommended (as the comment said in our code). I guess it is good for using with v2 also.

Note: I've explicitly enabled 200006 instead of 200001. Maybe we want to add a flag for this, but I think it is the better option.

README.md
```diff
@@ -235,11 +235,15 @@ All these variables impact in configuration directives in the modsecurity engine
...
 | MODSEC_AUDIT_LOG_PARTS  | A string that defines which parts of each transaction are going to be recorded in the audit log (Default: `'ABIJDEFHZ'`). See [SecAuditLogParts](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#secauditlogparts) for the accepted values. |
+| MODSEC_AUDIT_LOG_RELEVANT_STATUS | A regular expression string that defines the http error codes that are relevant for audit logging (Default: `"^(?:5|4(?!04))"`). See [SecAuditLogRelevantStatus](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secauditlogrelevantstatus) |
 | MODSEC_AUDIT_LOG_TYPE  | A string indicating the type of audit logging mechanism to be used (Default: `Serial`). Accepted values: `Serial`, `Concurrent` (`HTTPS` works only on Nginx - v3). See [SecAuditLogType](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secauditlogtype) for additional information. |
+| MODSEC_COOKIE_FORMAT | The cookie format used (Default: `0`)  :warning: Do not touch unless you really know what you are doing. |
```
theseion, line 246: Link to docs?

README.md
```diff
@@ -264,7 +268,10 @@ All these variables impact in configuration directives in the modsecurity engine
 | MODSEC_TAG  | A string indicating the default tag action, which will be inherited by the rules in the same configuration context (Default: `modsecurity`) |
 | MODSEC_TMP_DIR  | A string indicating the path where temporary files will be created (Default: `/tmp/modsecurity/tmp`) |
 | MODSEC_TMP_SAVE_UPLOADED_FILES  | A string indicating if temporary uploaded files are saved (Default: `On`) (only relevant in Apache - ModSecurity v2) |
+| MODSEC_UNICODE_MAPPING | The unicode Code Point to use form the default file(Default: `20127`). See [SecUnicodeMapFile](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secunicodemapfile) |
```
theseion, line 271:
```suggestion
| MODSEC_UNICODE_MAPPING | Unicode code point to use (Default: `20127`). See [SecUnicodeMapFile](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secunicodemapfile) |
```

README.md
```diff
@@ -264,7 +268,10 @@ All these variables impact in configuration directives in the modsecurity engine
...
 | MODSEC_UPLOAD_DIR  | A string indicating the path where intercepted files will be stored (Default: `/tmp/modsecurity/upload`) |
+| MODSEC_UPLOAD_FILE_MODE | (Default: `0600`) |
+| MODSEC_UPLOAD_KEEP_FILES | Configures whether or not the intercepted files will be kept after transaction is processed.  (Default: `RelevantOnly`) Accepted values: `On`, `Off`, `RelevantOnly`. See [SecUploadKeepFiles](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secuploadkeepfiles)
```
theseion, line 274:
```suggestion
| MODSEC_UPLOAD_KEEP_FILES | Configures whether the intercepted files will be kept after transaction is processed.  (Default: `RelevantOnly`) Accepted values: `On`, `Off`, `RelevantOnly`. See [SecUploadKeepFiles](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secuploadkeepfiles)
```

src/etc/modsecurity.d/modsecurity.conf
theseion: I agree.

theseion (changes requested):

README.md
fzipi: Added.

fzipi:
No comments from @dune73 or @airween. Will merge by EOD.

ruipin:
This pull request broke my setup using `nginx-alpine`.

To fix it I had to force the following values inside my docker compose:

```
      MODSEC_AUDIT_STORAGE_DIR: /var/log/modsecurity/audit/
      MODSEC_DISABLE_BACKEND_COMPRESSION: 'Off'
      MODSEC_UPLOAD_KEEP_FILES: 'Off'
```

The first one seems due to a typo - `MODSEC_AUDIT_STORAGE_DIR` is used by the new file, but the environment variable that is used for the default value is `MODSEC_AUDIT_STORAGE`.

The second one doesn't seem to have a default value.

The third one would complain that `RelevantOnly` was not a valid value, and it needed `On` or `Off`.

PR #245, opened 2024-05-06
fix: renovate update openresty

fzipi:
- openresty can have versions with up to 4 numbers
- it doesn't depend on nginx version


README.md
```diff
@@ -47,7 +47,7 @@ Examples:
 * nginx – *latest stable ModSecurity v3 on Nginx 1.25.3 official stable base image, and latest stable OWASP CRS 4.1.0*
    * [nginx](https://github.com/coreruleset/modsecurity-crs-docker/blob/master/nginx/Dockerfile)
    * [nginx-alpine](https://github.com/coreruleset/modsecurity-crs-docker/blob/master/nginx/Dockerfile-alpine)
-* Openresty - *last stable ModSecurity v3 on Nginx 1.25.3 official stable base image, and latest stable OWASP CRS 4.1.0*
+* Openresty - *last stable ModSecurity v3 on OpenResty 1.25.3 official stable base image, and latest stable OWASP CRS 4.1.0*
```
theseion, line 50:
```suggestion
* Openresty - *last stable ModSecurity v3 on OpenResty 1.25.3.1 official stable base image, and latest stable OWASP CRS 4.1.0*
```

docker-bake.hcl
```diff
@@ -179,7 +179,6 @@ target "openresty-alpine-fat" {
     dockerfile="openresty/Dockerfile-alpine"
     args = {
         OPENRESTY_VERSION = "${openresty-version}"
-        NGINX_VERSION = "${nginx-version}"
```
theseion, line 182: Can you derive the nginx version here instead?

theseion: And then also add `NGINX_VERSION` as `ARG` to the Dockerfile.

theseion (changes requested):

docker-bake.hcl
```diff
@@ -179,7 +179,7 @@ target "openresty-alpine-fat" {
...
+        NGINX_VERSION = patch(openresty-version)
```
theseion, line 182: Won't this just be `3` if openresty version is `1.25.3.1`?

fzipi: Nope, see https://github.com/coreruleset/modsecurity-crs-docker/blob/5039d9dce5689087d4d0b032476cb671f3e3913e/docker-bake.hcl#L83.

We get all the numbers up to the patch.

PR #253, opened 2024-05-09
fix: add missing JSON Content-Type rule

### LINKED ISSUE #252: modsecurity.conf does not create the same config as before
With #239, ModSecurity isn't configured the same way as it was before. This breaks the CRS test setup for example.

theseion:
The new modsecurity.conf file was missing rule 200001, which enables body processing of `application/json`.

Also took the opportunity to put the content of the file into a form that is easier to maintain w.r.t. to changes from upstream.

Fixes #252

### DIFF
--- a/src/etc/modsecurity.d/modsecurity.conf
+++ b/src/etc/modsecurity.d/modsecurity.conf
@@ -1,53 +1,52 @@
 # Original of the latest recommended version:
 # https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/modsecurity.conf-recommended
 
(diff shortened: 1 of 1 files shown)


fzipi:
Hmmmm.... https://github.com/coreruleset/modsecurity-crs-docker/pull/239#discussion_r1581753064 ?

fzipi:
Ok, so both are needed, right?

theseion:
Yes. One is for `application/json`, the other one for `application/<something>+json`.

fzipi (approved):

PR #254, opened 2024-05-12
Fix heredoc in check-low-port

woefe:
The container did not start anymore on the latest build. The logs didn't really tell what was going on, because the message from the check-low-port script is not printed. Heredoc needs to go through `cat`, not `echo`.


theseion (approved):
Thanks!

Issue #256, opened 2024-05-21
RESTRICTED_HEADERS CRS env variable not working with 202405101205 / CRS 4.2.0


combine-space:
After switching over to the latest release `202405101205`, the environment variable `RESTRICTED_HEADERS=` is not getting applied anymore. In CRS until v.4.0.0-rc1 the setvar name was `tx.restricted_headers` (https://github.com/coreruleset/coreruleset/blob/v4.0.0-rc1/crs-setup.conf.example#L492), in CRS v4.2.0 the rule has been splitted into two new rules, `tx.restricted_headers_basic` and `tx.restricted_headers_extended` (https://github.com/coreruleset/coreruleset/blob/v4.2.0/crs-setup.conf.example#L591-L625).

It seems, that the entrypoint script was not updated: https://github.com/coreruleset/modsecurity-crs-docker/blob/release/20240510/src/opt/modsecurity/activate-rules.sh#L62-L65

When running the sed line below with the corrected setvar name, it works again:
```
sed -z -E -i 's|#SecAction[^"]+"id:900250.*\x27tx\.restricted_headers_basic=[[:lower:][:space:]/-]*\x27\"|SecAction \\\n  \"id:900250, \\\n   phase:1, \\\n   nolog, \\\n   pass, \\\n   t:none, \\\n   setvar:\x27tx.restricted_headers_basic='"${RESTRICTED_HEADERS}"'\x27\"|' /etc/modsecurity.d/owasp-crs/crs-setup.conf
```

fzipi:
Thanks for the detailed report! Are you up for a patch?

fzipi:
Today's release should have your fix. Please test after baking have finished and let us know!

PR #257, opened 2024-05-22
fix: split restricted headers to work with crs 4.2

combine-space:
Adresses issue: https://github.com/coreruleset/modsecurity-crs-docker/issues/256


fzipi (approved):
Thanks for your contribution!

Changes look good, let's see why the pipeline didn't fired up.

fzipi:
Looks like we need #258 first.

fzipi:
Ok, needed to close and reopen, but looks like tests fired now.

PR #296, opened 2024-10-12
feat: Add Trivy scanner to the pipeline

huberts90:
Trivy action scans a Docker image against vulnerabilities in the CI/CD. The vulnerabilities cache is scheduled to be updated on a daily basis.


.github/workflows/verifyimage.yml
```diff
@@ -60,6 +65,18 @@ jobs:
...
 
+      - name: Scan ${{ matrix.target }}
+        uses: aquasecurity/trivy-action@0.27.0
+        with:
+          image-ref: ${{ needs.prepare.outputs.repo }}:${{ matrix.target }}
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
+          vuln-type: 'os,library'
+          severity: 'CRITICAL,HIGH,MEDIUM'
```
fzipi, line 76: I would say let's start with critical and highs.

huberts90: Sure, the scope has been narrowed down.

fzipi:
Nice, I think it has a good find. Will fix and then we can rebase.

fzipi:
@huberts90 Can you take a look on why this is failing now?

theseion:
I can't see the logs anymore, but, IIRC, the rate limit is tied to downloading the database and can be circumvented by using a different registry / mirror. GHCR would make sense, IMO.

fzipi:
Picked this up in #459 — thanks for the original work here, @huberts90, you're credited on the commit.

The rate limiting that blocked this is gone: Trivy pulls its database from `mirror.gcr.io` now with a `ghcr.io` fallback, the workflow already authenticates to ghcr, and the action caches the database. #459 also adds a daily scan of the published images that reports into code scanning, since a PR-time scan can't see vulnerabilities disclosed after a release.

fzipi:
Superseded by #459

PR #304, opened 2024-12-01
feat: add cors headers default nginx

### LINKED ISSUE #112: Adding CORS header 'Access-Control-Allow-Origin' with errors.
Hi

I  am using modsecurity-crs:nginx as a proxy for my backend, which is an API.

On a totally different machine, there is a frontend JS application. Modern JS needs to have a CORS header set.

The proxy and API work fine, when everything is OK, but when a rule triggers the CORS header is not added.

----

Example:

The API allows users to POST data - up to 25MB. Currently the application checks the file size and also the proxy checks it.

When the proxy intercepts a max_body_size (or similar variable) that is larger than allowed it sends a 403 request, that does not include the CORS header. On the other hand the application sets it, when it encounters that denial.

I guess the easy way is for the application to block it, but it's better if it is intercepted earlier?

EDIT:

Currently my **default.conf.template** that I map to the container contains the following (which does not work):

```nginx
# Nginx configuration for both HTTP and SSL

map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}

# https://serverfault.com/questions/958965/nginx-enabling-cors-for-multiple-subdomains

map $http_origin $allow_origin {
    ~^https?://(.*\.)?.mydomain.example(:\d+)?$ $http_origin;
    ~^https?://(.*\.)?localhost(:\d+)?$ $http_origin;
    default "";
}

server {
# http redirect
}

server {
# stuff
    add_header 'Access-Control-Allow-Origin' $allow_origin;
    location / {
# stuff
    }
}
```

fzipi:
## what

- add cors hedaers to the 403 answer

## why

Fixes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,10 +208,10 @@ These variables are common to image variants and will set defaults based on the 
 
 | Name     | Description|
 | -------- | ------------------------------------------------------------------- |
-| CORS_HEADER_403_ALLOW_ORIGIN | When generating a 403 response from the server, the origin can access this response using the [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) header (Default: `"*"`). |
-| CORS_HEADER_403_ALLOW_METHODS | When generating a 403 response from the server, the methods that will be sent in the header [Access-Control-Request-Method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Method) . Default: `"GET, POST, PUT, DELETE, OPTIONS"` |
-| CORS_HEADER_403_CONTENT_TYPE | When generating a 403 response from the server, the content-type of the response. Default: (`"text/plain"`) |
-| CORS_HEADER_403_MAX_AGE | When generating a 403 response from the server, the content of the [Access-Control-Max-Age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age) header. (Default: `3600`) |
+| CORS_HEADER_403_ALLOW_ORIGIN | The value of the [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) header for `403` responses. Determines which origins can access the response. (Default: `"*"`). |
+| CORS_HEADER_403_ALLOW_METHODS | The value of the [Access-Control-Request-Method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Method) header for `403` responses. Determines the allowed request methods for the resource. Default: `"GET, POST, PUT, DELETE, OPTIONS"` |
+| CORS_HEADER_403_CONTENT_TYPE | The value of the  `Content-Type` header for `403` responses. Default: (`"text/plain"`) |
+| CORS_HEADER_403_MAX_AGE | The value of the [Access-Control-Max-Age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age) header for `403` responses. The number of seconds that preflight requests for this resource may be cached by the browser. (Default: `3600`) |
 | DNS_SERVER  | A string indicating the name servers used to resolve names of upstream servers into addresses. For localhost backend this value should not be defined (Default: _not defined_) |
 | KEEPALIVE_TIMEOUT  | Number of seconds for a keep-alive client connection to stay open on the server side (Default: `60s`) |
 | NGINX_ALWAYS_TLS_REDIRECT | A string value indicating if http should redirect to https (Allowed values: `on`, `off`. Default: `off`) |
--- a/README.md
+++ b/README.md
@@ -208,6 +208,10 @@ These variables are common to image variants and will set defaults based on the 
 
 | Name     | Description|
 | -------- | ------------------------------------------------------------------- |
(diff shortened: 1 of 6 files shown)


nginx/templates/includes/cors.conf.template
```diff
@@ -0,0 +1,5 @@
+more_set_headers -s 403 'Content-Type' '${CORS_CONTENT_TYPE}';
```
theseion, line 1: The values should all be expanded from variables, I presume.

theseion (changes requested):

nginx/templates/includes/cors.conf.template
theseion: Yes, but you're only replacing a single variable. You're not using the ones you've defined. And `CORS_CONTENT_TYPE` doesn't exist at all.

fzipi: WTH. Bad version. 🤦 

Don't know if I can write a decent test for this 🤔

nginx/templates/includes/cors.conf.template
```diff
@@ -0,0 +1,5 @@
+more_set_headers -s 403 'Content-Type' '${CORS_HEADER_403_CONTENT_TYPE}';
+more_set_headers -s 403 'Access-Control-Allow-Origin' '${CORS_HEADER_403_ALLOW_ORIGIN}';
+more_set_headers -s 403 'Access-Control-Max-Age' '${CORS_HEADER_403_MAX_AGE}';
+more_set_headers -s 403 'Access-Control-Allow-Methods' '${CORS_HEADER_403_ALLOW_METHODS}';
+more_set_headers 'Access-Control-Allow-Headers' '${CORS_HEADER_ACCESS_CONTROL_ALLOW_HEADERS}';
```
theseion, line 5: I believe the default value should be "unset". According to the documentation, there's a list of headers that are allowed by default. In many cases, that list should not need to be extended. By supplying `*` as the value, we would allow _any_ header.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

theseion (changes requested):

nginx/Dockerfile
```diff
@@ -114,6 +114,11 @@ LABEL maintainer="Felipe Zipitria <felipe.zipitria@owasp.org>"
 ENV \
     ACCESSLOG=/var/log/nginx/access.log \
     BACKEND=http://localhost:80 \
+    CORS_HEADER_403_ALLOW_ORIGIN="*" \
+    CORS_HEADER_403_ALLOW_METHODS="GET, POST, PUT, DELETE, OPTIONS" \
+    CORS_HEADER_403_CONTENT_TYPE="text/plain" \
+    CORS_HEADER_403_MAX_AGE=3600 \
+    CORS_HEADER_ACCESS_CONTROL_ALLOW_HEADERS="*" \
```
fzipi, line 121:
```suggestion
```

nginx/Dockerfile-alpine
```diff
@@ -107,6 +107,11 @@ LABEL maintainer="Felipe Zipitria <felipe.zipitria@owasp.org>"
 ENV \
     ACCESSLOG=/var/log/nginx/access.log \
     BACKEND=http://localhost:80 \
+    CORS_HEADER_403_ALLOW_ORIGIN="*" \
+    CORS_HEADER_403_ALLOW_METHODS="GET, POST, PUT, DELETE, OPTIONS" \
+    CORS_HEADER_403_CONTENT_TYPE="text/plain" \
+    CORS_HEADER_403_MAX_AGE=3600 \
+    CORS_HEADER_ACCESS_CONTROL_ALLOW_HEADERS="*" \
```
fzipi, line 114:
```suggestion
```

nginx/templates/includes/cors.conf.template
fzipi, line 5:
```suggestion
```

theseion (approved):

README.md
```diff
@@ -208,6 +208,10 @@ These variables are common to image variants and will set defaults based on the
 
 | Name     | Description|
 | -------- | ------------------------------------------------------------------- |
+| CORS_HEADER_403_ALLOW_ORIGIN | When generating a 403 response from the server, the origin can access this response using the [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) header (Default: `"*"`). |
+| CORS_HEADER_403_ALLOW_METHODS | When generating a 403 response from the server, the methods that will be sent in the header [Access-Control-Request-Method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Method) . Default: `"GET, POST, PUT, DELETE, OPTIONS"` |
+| CORS_HEADER_403_CONTENT_TYPE | When generating a 403 response from the server, the content-type of the response. Default: (`"text/plain"`) |
+| CORS_HEADER_403_MAX_AGE | When generating a 403 response from the server, the content of the [Access-Control-Max-Age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age) header. (Default: `3600`) |
```
theseion, line 214:
```suggestion
| CORS_HEADER_403_MAX_AGE | The value of the [Access-Control-Max-Age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age) header for `403` responses. The number of seconds that preflight requests for this resource may be cached by the browser. (Default: `3600`) |
```

theseion (dismissed):

theseion (approved):

PR #314, opened 2024-12-22
feat: improve rule configuration

theseion:
- Configure rules to update in a separate configuration file
- Use generic script to configure rules that doesn't rely on exact match of rule
- Log which rules are configured
- Log when a rule that should be configured wasn't found


fzipi (dismissed):
Well, it is difficult to have tests for this I guess.

theseion:
@fzipi I've implemented a simple way to test the logic. That did uncover a few issues with the code I had written.

fzipi (approved):
LGTM!

PR #318, opened 2025-01-06
docs: fix default value for ALLOWED_REQUEST_CONTENT_TYPE

andreasferber:
The default value for ALLOWED_REQUEST_CONTENT_TYPE shown in README.md doesn't match the actual default value in CRS, which is misleading if you try to customize the setting by following the example. This PR corrects that.


fzipi (approved):
Of course, the pipe is also used as column separator. Thanks for the fix!

PR #328, opened 2025-02-01
use stable version of modsecurity nginx connector

### LINKED ISSUE #326: latest nginx-alpine tag is not working properly
in the following example this is a small docker stack with owasp/modsecurity-crs as proxy and portainer behind owasp/modsecurity-crs

the latest tag owasp/modsecurity-crs:4.10.0-nginx-alpine-202501270601 produces weird errors

Image

the previous tag owasp/modsecurity-crs:4.10.0-nginx-alpine-202501050801 works perfectly

Image

it looks suspicious to me. both image are from the public docker registry. i dont have the time to debug this. i can assure that the config is exactly the same.

fichte:
this fixes #326

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -24,6 +24,10 @@ variable "httpd-version" {
     default = "2.4.63"
 }
 
+variable "modsecurity-nginx-version" {
+    default = "1.0.3"
+}
+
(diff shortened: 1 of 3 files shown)


fzipi (approved):
Let's fix this first, then we optimise.

PR #331, opened 2025-02-06
fix: replace `$upstream` variable with `BACKEND` env var

### LINKED ISSUE #330: stop using $upstream variable proxy_backend.conf.template for owasp/modsecurity-crs:nginx
What's the original intent on using $upstream in the proxy_backend.conf.template? If nginx has to replace proxy_pass argument with a variable, it will end up using the DNS resolver to resolve the hostname, and not /etc/hosts first, which defeats the purpose of extra_hosts. Because it's a template, can we replace the proxy_pass argument with the BACKEND environment variable instead? I had to manually mount an edited template to force nginx to resolve a hostname mapped through extra_hosts.

theseion:
`proxy_pass` will use the resolver whenever the argument contains variables. This isn't usually an issue but it can be if hosts are declared via `/etc/hosts`, e.g. when useing `extra_hosts` from docker compose. Then, `/etc/hosts` would not be considered.

Use the `BACKEND` variable directly to avoid use of the resolver.

Fixes #330

### DIFF
--- a/nginx/templates/conf.d/default.conf.template
+++ b/nginx/templates/conf.d/default.conf.template
@@ -11,7 +11,6 @@ server {
     listen ${PORT} default_server;
 
     server_name ${SERVER_NAME};
-    set $upstream ${BACKEND};
     set $always_redirect ${NGINX_ALWAYS_TLS_REDIRECT};
 
     PROXY_SSL_CONFIG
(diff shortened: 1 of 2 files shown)


fzipi (approved):

gtaws:
Thanks so much! Just to elaborate (because it took me a real long time on Google to figure this out), if proxy_pass sees a variable as its parameter, it will no longer consider /etc/hosts when resolving a host, and only refer the DNS resolver it's been configured to resolve hosts. If proxy_pass sees a literal as its parameter, it will consider /etc/hosts first before referring to the DNS resolver. This is important if you depend on extra_hosts for your container to function correctly.

PR #339, opened 2025-02-26
fix: default paranoia settings break CRS

theseion:
We are still setting `PARANOIA` as default in some places, which causes detection and blocking levels to be in an invalid state (detection must be >= blocking).

This PR also adds running CRS tests to the verification, as we've run into issues multiple times that we would have caught, had we run the test suite (this issue in particular).


theseion:
@fzipi I want to merge this PR and then release.

theseion:
Well bummer. I'll need to investigate those issues, but tonight I won't be around. It will have to wait until Thursday evening probably.

theseion:
@fzipi Ready. I had to disable alpine and read-only builds. The alpine images have had issues for a while, I think.

.github/workflows/verifyimage.yml
```diff
@@ -107,3 +108,41 @@ jobs:
...
+            's#image: owasp/modsecurity-crs:${{ contains(matrix.target, 'apache') && 'apache' || 'nginx' }}.*#image: ${{ matrix.target }}-verification#' \
+            crs/tests/docker-compose.yml
+      - name: Run CRS tests for ${{ matrix.target }}
+        run: |
+          cd crs
+          modsec_version="${{ contains(matrix.target, 'apache') && 'modsec2-apache' || 'modsec3-nginx' }}"
+          backend="${{ contains(matrix.target, 'apache') && 'httpd' || 'nginx' }}"
+          mkdir -p "tests/logs/${modsec_version}/{nginx,apache2}"
+          docker compose -f ./tests/docker-compose.yml up -d "${modsec_version}"
+          docker compose -f ./tests/docker-compose.yml logs
```
nakamo326, line 137: Perhaps the reason the go-ftw test is failing with the alpine image is that the test begins before nginx has started. 
It might be fixed to wait for nginx to start by using sleep etc.

```suggestion
          docker compose -f ./tests/docker-compose.yml up -d "${modsec_version}"
          sleep 30
          docker compose -f ./tests/docker-compose.yml logs
```

theseion: `up` is synchronous. It shouldn't be necessary to wait.

theseion: It's definitely a segfault.

PR #364, opened 2025-07-22
Allow IPv6 resolver for nginx

ramann:
Check if DNS_SERVER contains colons (IPv6) and wrap in brackets for nginx


theseion:
Thanks for this PR @ramann. It wouldn't have worked though for manually specified IPv6 addresses already in the correct format. Also, it would have broken with more than one address. I've created a PR that replaces this one and improves on the configuration of the `resolver` directive: #365.

PR #372, opened 2025-08-28
feat: improve HTTP/2 configuration

theseion:
- add explicit HTTP/2 configuration to nginx
- add `h2c` to default protocols for httpd
- add `H2Direct` directive and configuration option to httpd

### DIFF
--- a/.github/workflows/verifyimage.yml
+++ b/.github/workflows/verifyimage.yml
@@ -88,26 +88,89 @@ jobs:
             README.md \
             "${{ matrix.target }}.env"
           echo "Starting container ${{ matrix.target }}-verification"
+          # The backend is a dummy
           docker run \
             --pull "never" \
             -d \
             --name ${{ matrix.target }}-test \
             --env-file "${{ matrix.target }}.env" \
+            -p 8080:8080 \
+            -e BACKEND=http://localhost:9999 \
             "${{ matrix.target }}-verification"
-          sleep 30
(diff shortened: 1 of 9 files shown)


fzipi:
Do we want to add a test [here](https://github.com/coreruleset/modsecurity-crs-docker/blob/main/.github/workflows/verifyimage.yml#L102-L110) like calling `curl --http2 --verbose "http://localhost" 2>&1 | grep 'using HTTP/2'`?

.github/workflows/verifyimage.yml
```diff
@@ -99,16 +99,54 @@ jobs:
...
-            curl -q -D headers.txt http://localhost:8080/?test=../../etc/passwd
+            echo "### nginx tests ###"
+
+            echo "Storing headers for sample attack..."
+            curl -q -D headers.txt "${HOST_AND_PORT}/?test=../../etc/passwd"
+            echo "...done"
+
+            echo "Check status 403..."
             grep -q "HTTP/1.1 403 Forbidden" headers.txt
+            echo "...yes"
```
fzipi, line 113: Maybe you want to combine `echo -e` with echo.
Or use `printf` in the first line without `\n` and then just doing `echo "yes"`, so it will show better in one line.

theseion:
Yeah, sorry, just broke something and am trying to fix it again :)

theseion:
ok, fixed now.

PR #390, opened 2025-11-01
feat: Nginx configurable X-Forwarded-Proto

xx4h:
Make `X-Forwarded-Proto` configurable for proxy chain scenario with only `https` on first/initial proxy.


fzipi (approved):
LGTM

PR #428, opened 2026-05-19
ci: use arm runners when available

fzipi:
## what

- use arm runners when building the arm version

## why

- GH has support for ARM runners now
- Check more platforms when building

## Release Notes

* **Chores**
  * Improved Docker multi-platform build and verification workflows by propagating selected platforms through the Buildx/Bake matrix, ensuring build/push targets are generated for the intended architectures.
  * Updated runner selection to automatically use ARM-compatible runners when building ARM platforms.
  * Added configuration integrity checks during the workflow and adjusted verification/testing downloads to match the target architecture.

### DIFF
--- a/.github/workflows/verifyimage.yml
+++ b/.github/workflows/verifyimage.yml
@@ -196,8 +196,9 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           cd crs
+          ARCH="${{ startsWith(matrix.platforms, 'linux/arm') && 'arm64' || 'amd64' }}"
           gh release download -R coreruleset/go-ftw "v${{ env.GO_FTW_VERSION }}" \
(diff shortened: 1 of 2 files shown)


.github/workflows/verifyimage.yml
```diff
@@ -73,10 +74,10 @@ jobs:
           files: |
             ./docker-bake.hcl
           targets: ${{ matrix.target }}
-          # Build only linux/amd64 and tag the images as verification builds.
```
theseion, line 76: We had specifically restricted verification to linux/amd64 to minimize build time. With this change, we would again build all images during verification. Do we really want to do that? I'm not against it per se, but it takes a very long time to build all the images.

fzipi: Well... technically that was because qemu emulates the aarch (arm). With this it should use native runners, and cut times to be equivalent to x86.

Still needs to be tested.

fzipi:
Well... nice! Under 2 minutes! 🎉

fzipi:
@theseion can you reapprove?

theseion (approved):

Issue #443, opened 2026-07-03
Images only available for single architecture.


pfandzelter:
Hello, thank you for your work!
Unless I misunderstand how/when LTS images are available, it appears that a recent change has led to LTS to being only pushed for `linux/arm64` (Apache) and `linux/386` (NGINX). See Docker Hub:
Nginx: https://hub.docker.com/r/owasp/modsecurity-crs/tags?name=4.25-nginx
Apache: https://hub.docker.com/r/owasp/modsecurity-crs/tags?name=4.25-apache

I would expect these images to remain multi-architecture.
Thanks!

fzipi:
@pfandzelter that's weird. Do you have time for debugging this?

pfandzelter:
I assume that this was introduced with 7c1554f. CI for the last release (https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/28645668048) shows an individual job for each platform, while the previous (https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/28404527510) had all three platforms within a job. That will cause some race where a push to a tag overwrites a previous tag and only the last architecture that was built remains on Docker Hub

fzipi:
Thanks! Care to send a PR with the fix? 😃

pfandzelter:
I could look into using the [docker/github-builder/bake.yml](https://docs.docker.com/build/ci/github-actions/multi-platform/#with-bake) action, would that work for you? Simplifies the CI but introduces a dependency on that action 🤔 Manually extending the current publish.yml would probably (?) involve pushing each image by digest (not tag) and then tagging the multi-arch manifest in a separate step. Or going back to the non-parallelized build, which isnt optimal

fzipi:
I think we had this problem in the past. Whatever you think is good, give it a try. We are open to iterate.

fzipi:
Root cause confirmed: 7c1554f split each bake target's build into one job per platform (to use native ARM runners), but every job still pushed under the same final tags. The three concurrent pushes for a target raced, and only the last platform to finish survived on Docker Hub/GHCR — that's why LTS tags ended up single-arch.

#445 fixes this by reverting to one job per target (all platforms built together via QEMU), which is correct but gives up the native-ARM speedup.

[...]

PR #445, opened 2026-07-03
fix: don't split image build by target platform

pfandzelter:
As discussed in #443, 7c1554f introduced parallelized image builds across bake targets **and** platforms (previously it was just by bake targets). This introduced an error where multiple image for the same target (i.e., tag) but with different platforms were built and pushed concurrently and only the final push remained on Docker Hub/GHCR, so each tag had only one platform available.

Fixing this in the current setup would require pushing individual digests first and then tagging them at the end by manually building a manifest, as [docker/github-builder](https://github.com/docker/github-builder/blob/2ed0b42d4b91658ed511ca56581527d0af1f958d/.github/workflows/bake.yml#L1173) does. This is likely difficult to support as it is complex and requires duplicating image names to the GitHub CI setup.

I tried setting up the docker/github-builder action for this repository but it conflicts with the extensive tag setup in the Bakefile (the action requires meta-images for each additional tag).

I thus simply reverted the changes to .github/workflows/publish.yml (.github/workflows/verifyimage.yml works correctly as far as I can see from [this job](https://github.com/coreruleset/modsecurity-crs-docker/actions/runs/28405722161)). Unfortunately the CI pipeline is now much slower (as it was before 7c1554f) because of cross-platform builds and no parallelization.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,11 +21,10 @@ jobs:
         uses: docker/bake-action/subaction/matrix@5be5f02ff8819ecd3092ea6b2e6261c31774f2b4 # v6
         with:
           target: default
-          fields: platforms
 
   build:
     name: Build ${{ matrix.target }}
-    runs-on: ${{ startsWith(matrix.platforms, 'linux/arm') && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
+    runs-on: ubuntu-latest
     needs:
       - prepare
     permissions:
@@ -76,8 +75,6 @@ jobs:
           files: |
             ./docker-bake.hcl
           targets: ${{ matrix.target }}
-          set: |
-            *.platform=${{ matrix.platforms }}
           push: true
           provenance: true
           sbom: true


fzipi:
This is on me, need to sign the release. Give me a sec.

fzipi:
Thanks for tracking this down and for the writeup on the digest+merge alternative you considered — that's exactly the direction I went in #447.

Instead of using docker/github-builder's bake.yml (which needs meta-images/meta-tags to drive tagging via docker/metadata-action), #447 keeps docker-bake.hcl untouched and reuses its already-resolved `tags` output directly (`docker buildx bake --print`, captured once in `prepare` and shared via artifact so the embedded timestamp doesn't diverge across jobs). Each per-platform job pushes by digest only, and a new merge job runs `docker buildx imagetools create` per target to assemble the manifest and push to both registries.

This keeps the native-ARM-runner speedup from 7c1554f instead of giving it up. I validated the digest-push + cross-registry imagetools create mechanics locally against two throwaway registries before opening it — details in the PR description. Leaving this open for you/@fzipi to compare against #447 and decide which one to merge.

fzipi:
@pfandzelter Can you check if the latest release works for you? My tests say it is fixed.

pfandzelter:
@fzipi Thank you, it works as expected now. 😃  I'll close this then

PR #446, opened 2026-07-03
feat: auto-release on tracked upstream version bump

### LINKED ISSUE #440: Release when upstream dependencies upgrade
We should create a release automatically when any of our upstream dependencies upgrade.

That way we can move faster with releases.

It should only release automatically when:
- modsecurity2
- libmodsecurity3 
- nginxinc/nginx-unprivileged
- httpd
get updated by renovatebot.

fzipi:
## Summary
- Adds `.github/workflows/auto-release.yml`, triggered on push to `main` when `docker-bake.hcl` changes
- Compares the `modsec2-version`, `modsec3-version`, `httpd-version`, and `nginx-version` values before/after the push; if any changed, it creates a `release/YYYYMMDD` tag (with a numeric suffix on same-day collisions) and a GitHub release
- Adds `workflow_call` (with explicit `dockerhub_user`/`dockerhub_token` secret inputs) to `publish.yml` so the new workflow can invoke the build-and-push job directly, without depending on `GITHUB_TOKEN`-created releases re-triggering the `release: published` event (which GitHub Actions does not do)

## Test plan
- [ ] `actionlint` / `zizmor` pass on both workflow files
- [ ] Merge a PR that bumps one of the four tracked versions in `docker-bake.hcl` and confirm a release is tagged and images are published
- [ ] Merge a PR that changes `docker-bake.hcl` without touching the tracked versions (e.g. a CRS bump) and confirm no release is created

## AI Disclosure

2. **What was generated or assisted**: Design and implementation of `.github/workflows/auto-release.yml` (diffing tracked version variables in `docker-bake.hcl` between pushes, tagging/creating the release with same-day collision handling) and the `workflow_call` addition to `publish.yml` with explicit secret inputs, needed because GitHub Actions doesn't let a `GITHUB_TOKEN`-created release re-trigger the `release: published` event.
3. **Review performed**: `actionlint` and `zizmor` run against both workflow files (zero findings). I reviewed the job/permission structure and the version-diff logic myself before publishing; the actual tag-and-publish path is still marked as unverified in the test plan below pending a real merge that bumps one of the tracked versions.

## Summary by CodeRabbit

* **New Features**
  * Releases are now created automatically when monitored upstream versions change.
  * Release notes are generated automatically, and publishing begins without manual intervention.
  * Date-based release tags provide consistent identifiers for each published version.

* **Chores**
  * Automated publishing now requires Docker Hub credentials, helping ensure releases are securely built and delivered.
  * Publishing workflows now use scoped permissions and organized run handling.


.github/workflows/auto-release.yml
```diff
@@ -0,0 +1,97 @@
...
+          bumped=false
+          for var in modsec2-version modsec3-version httpd-version nginx-version; do
+            old=$(version_of "$BEFORE" "$var")
+            new=$(version_of "${{ github.sha }}" "$var")
+            if [ "$old" != "$new" ]; then
+              echo "Version bump detected for $var: $old -> $new"
+              bumped=true
+            fi
+          done
+          echo "bumped=$bumped" >> "$GITHUB_OUTPUT"
```
theseion:
```suggestion
          set -euo pipefail
          old_sha="${{ github.event.before }}"
          if [ -z "${old_sha//0/}" ]; then
            echo "No previous commit to diff against; skipping."
            echo "bumped=false" >> "${GITHUB_OUTPUT}"
            exit 0
          fi

          version_of() {
            sha=${1}
            variant=${2}
            git show "${sha}:docker-bake.hcl" | awk -v variant="${variant}" '
              $0 ~ "variable \"" variant "\"" { found=1 }
              found && /default *=/ { print; exit }
            '
          }

          bumped=false
          for variant in modsec2-version modsec3-version httpd-version nginx-version; do
            old=$(version_of "${old_sha}" "${variant}")
            new=$(version_of "${{ github.sha }}" "${variant}")
            if [ "${old}" != "${new}" ]; then
              echo "Version bump detected for $“variant}: ${old} -> ${new}"
              bumped=true
            fi
          done
          echo "bumped=${bumped}" >> "${GITHUB_OUTPUT}"
```

.github/workflows/auto-release.yml
```diff
@@ -0,0 +1,97 @@
...
+          suffix=1
+          while gh api "repos/${{ github.repository }}/git/ref/tags/$tag" >/dev/null 2>&1; do
+            tag="${base}-${suffix}"
+            suffix=$((suffix + 1))
+          done
+
+          gh release create "$tag" \
+            --target "${{ github.sha }}" \
+            --title "$tag" \
+            --generate-notes
```
theseion, line 88:
```suggestion
          set -euo pipefail
          base="release/$(date -u +%Y%m%d)"
          tag="${base}"
          suffix=1
          while gh api "repos/${{ github.repository }}/git/ref/tags/${tag}" >/dev/null 2>&1; do
            tag="${base}-${suffix}"
            suffix=$((suffix + 1))
          done

          gh release create "${tag}" \
            --target "${{ github.sha }}" \
            --title "${tag}" \
            --generate-notes
```

theseion (changes requested):

fzipi:
Both review threads are applied on the branch, and I ran the detection logic against real history rather than leaving the test plan unchecked.

**@theseion's suggestions** are in — the `github.event.before` all-zeros guard and the `version_of` helper on line 56, and the tag collision loop on line 88. One fix on top: the suggestion carried a unicode quote, so the log line read `$“variant}` and printed a literal instead of the variable name. `shellcheck` flags it as SC1111 through `actionlint`, but nothing in CI lints workflows, so the checks stayed green. Fixed in 613c731.

**CodeRabbit's concurrency point** was addressed earlier in a95ae81 with the workflow-level `concurrency` group.

I did not add the second half of that comment, distinguishing a transient `gh api` failure from a genuine 404 in the tag-existence loop. If a request fails for another reason the script treats the name as free, `gh release create` then fails on the duplicate tag, and the run goes red — a loud failure rather than a wrong release, so the guard would only change the error message.

**Detection logic, replayed against real commits:**

| commit | change | result |
|---|---|---|
| e7f7808 | nginx 1.30.3 → 1.30.4 | `bumped=true` |
| 3751d16 | httpd 2.4.67 → 2.4.68 | `bumped=true` |
| 1a7599b | modsec3 3.0.15 → 3.0.16 and nginx 1.30.2 → 1.30.3 | `bumped=true`, both reported |
| b015e92 | CRS versions only | `bumped=false` |

That last row is the case the workflow must not fire on, and it is the third item in the test plan. The remaining unchecked item is the live one: merging a real version bump and watching a release get tagged and published, which can only be exercised after this merges.

theseion (approved):

PR #458, opened 2026-08-16
fix: CORS_HEADER_403_CONTENT_TYPE mismatch in Dockerfile-alpine

fzipi:
`nginx/Dockerfile-alpine` set `CORS_HEADER_403_CONTENT_TYPE` to `"text/plain"` while `nginx/Dockerfile` and the README both document the default as `"text/html"`, so the two image variants behaved differently.

## Change

- **`nginx/Dockerfile-alpine`**: `CORS_HEADER_403_CONTENT_TYPE` default `"text/plain"` → `"text/html"`, matching `nginx/Dockerfile` and the README.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the default content type for CORS-related 403 error responses from plain text to HTML.
  - Improved validation to ensure these responses consistently include the expected HTML content type.
  - Enhanced deployment verification across supported targets, helping detect configuration or startup issues more reliably before release.


.github/workflows/verifyimage.yml
```diff
@@ -137,12 +141,12 @@ jobs:
               sleep 5
             else
               printf "No more retries. Here's the log output from the container\n\n\n"
-              docker logs ${{ matrix.target }}-test
+              docker logs "${TARGET}-test"
               exit 1
             fi
           done
 
-          if grep -q  "nginx"<<< "${{ matrix.target }}"; then
```
theseion, line 145: We should run these tests as a dedicated test suite, not as part of the startup. Create a follow-up issue.

fzipi: Agreed. Filed #461 — it covers the assertions in this step plus the duplicated startup wait in the HTTP/3 job, and leaves the harness choice (plain script vs bats vs stretching go-ftw) open for discussion there.

Keeping the Content-Type assertion in this PR for now so the default it guards doesn't regress before that lands.

theseion (approved):

PR #459, opened 2026-08-16
feat(ci): scan images for vulnerabilities with Trivy

### LINKED ISSUE #6: Consider adding docker security scanner?
We may want to add an action for docker security scanning:

https://github.com/phonito/phonito-scanner-action

fzipi:
Takes over #296 by @huberts90, which stalled on Trivy database rate limiting.

## How it runs

**Pull requests** (`verifyimage.yml`): the build job scans the image it just built, as its last step so a vulnerable base image does not hide functional test results. It fails on `HIGH,CRITICAL` with `ignore-unfixed`, so only vulnerabilities the distribution has already patched block a merge — the fix is a base image bump, which Renovate raises. The unfixed CVEs Debian and Alpine carry never block.

**Daily** (`scan.yml`): scans the published `ghcr.io` images and uploads SARIF to code scanning rather than failing. Vulnerabilities disclosed after a release cannot be caught by a pull request scan, and the fix for them is cutting a release, not turning a scheduled run red. The image list is derived from `docker buildx bake --print`, one scan per floating tag (the CRS version variants of a target share a base image, so they share an OS package set).

## Why Trivy

For these images the results are decided by the Alpine `secdb` and Debian Security Tracker feeds, which Trivy consumes directly, and both distributions are first-class. Grype scores similarly on distribution packages but adds binary classifiers that guess versions for source-built binaries — this repository compiles ModSecurity, the nginx modules and the Lua modules from source, so those classifiers are a noise source here rather than added coverage. Docker Scout needs per-repository enrollment and would not run on fork pull requests.

The rate limiting that stalled #296 no longer applies: Trivy now pulls its database from `mirror.gcr.io` with a fallback to `ghcr.io`, both workflows authenticate to `ghcr.io`, and the action caches the database between runs.

## False positives

Configuration, not engine choice, decides the noise level here:

- `ignore-unfixed` — the dominant filter. Distributions mark large numbers of CVEs as unfixed or not-affected.
- `scanners: vuln` only. Secret scanning of a built image flags generated TLS material.
- `severity: HIGH,CRITICAL`.

Measured with these flags:

| image | fixable HIGH/CRITICAL |
|---|---|
| `nginxinc/nginx-unprivileged:1.30.4` (base) | 0 |
| `httpd:2.4.68` (base) | 0 |
| the debian runtime package set, freshly installed | 0 |
| `ghcr.io/coreruleset/modsecurity-crs:nginx-alpine` (published) | 0 |
| `ghcr.io/coreruleset/modsecurity-crs:apache-alpine` (published) | 4 |
| `ghcr.io/coreruleset/modsecurity-crs:nginx` (published) | 486 |
| `ghcr.io/coreruleset/modsecurity-crs:apache` (published) | 630 |

The gate is quiet on current builds. The published debian images carry a backlog because they were built at release time and Debian has patched since — 395 of the 486 findings on `:nginx` are `linux-libc-dev`, kernel headers that are in the image only because the runtime stage installs `libcurl4-gnutls-dev` and `libmaxminddb-dev` instead of their runtime packages. Replacing those two is a separate change; it removes most of the scheduled scan output along with the attack surface.

## Summary by CodeRabbit

- **Security**
  - Added automated vulnerability scanning for published container images.
  - Scans identify fixed high- and critical-severity operating system and library vulnerabilities.
  - Verification scans can block builds when qualifying vulnerabilities are detected.
  - Results are available through code-scanning reports.
  - Scans run automatically on a schedule and can also be started manually.
- **Maintenance**
  - Improved tracking and maintenance of security scanning tool versions.

Closes #6.


.github/workflows/scan.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0 # ratchet:docker/login-action@v4.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Reports instead of failing: a fix for a published image means cutting a release,
+      # not turning this run red. The findings land in the Security tab.
+      - name: Scan ${{ matrix.image }} (${{ matrix.platform }})
+        uses: aquasecurity/trivy-action@a9c7b0f06e461e9d4b4d1711f154ee024b8d7ab8 # ratchet:aquasecurity/trivy-action@v0.36.0
```
theseion: Don't do this. Trivy is (as we now know) a prime target for an attack. We should isolate it to have access to necessary resources only. I've implemented the following at work:
1. Use a trivy container image with a cache volume to cache the vulnerability database
2. Use a trivy container image and mount resources to scan into the container
    1. mount the cache
    2. Run the scan container with `--network none` to prevent an infected scanner from doing anything

theseion (changes requested):

.github/workflows/scan.yml
fzipi: Done in 58ff7ab — both workflows now run Trivy as a container with `--network none`, matching the three points.

```
docker run --rm --network none \
  -v "${PWD}/.trivy-cache:/cache" \
  -v "${PWD}/${TARGET}-verification.tar:/scan/image.tar:ro" \
  "${TRIVY_IMAGE}" image --cache-dir /cache --skip-db-update --input /scan/image.tar ...
```

- **cache volume**: `.trivy-cache` is mounted in and restored by `actions/cache`, keyed per day so the repository cache does not fill with copies
- **resources mounted in**: the pull request scan reads the tar `bake` already writes, so it never touches a registry; the scheduled scan does `docker pull --platform` and `docker save` first, then hands over the tar. That also retires the `TRIVY_PLATFORM` workaround, since Docker selects the platform now
- **`--network none`**: the only networked step is `--download-db-only`, in a separate step of its own

The image is pinned by digest (`aquasec/trivy:0.74.0@sha256:62b1e65e…`) with a Renovate comment, so the scanner itself cannot change under us either.

## What testing turned up

Running it offline the obvious way fails on the debian images:

```
FATAL  Unable to initialize the Java DB: '--skip-java-db-update' cannot be specified on the first run
```

The debian base ships `/usr/share/java/libintl*.jar` and `/usr/share/maven-repo`, so Trivy wants its Java database — 908MB — for jars that come from `gettext-base`, a package the OS scan already covers. Skipping those two directories avoids the download entirely without losing anything.

Verified locally with the pinned image:

| check | result |
|---|---|
| offline scan of a buildkit `type=docker,dest=` tar | works, correct target detected |
| offline scan of published `:nginx` | 9 fixable HIGH+CRITICAL, `CVE-2026-53615` on util-linux — same as the action reported |
| offline scan of published `:nginx-alpine` | 0, as before |
| SARIF through the mounted output directory | 9 results, `HIGH` only, readable by the runner afterwards |
| scan duration once the database is cached | about one second |

The severity filter applies to SARIF directly from the CLI, so `limit-severities-for-sarif` is gone as well.

fzipi:
Valid finding, fixed in 0e72b2f. The manager was left over from the trivy-action version and matched nothing at all after the switch to a container:

```
"# renovate: datasource=(?<datasource>\\S+) depName=(?<depName>\\S+)\\s+version: (?<currentValue>\\S+)"
```

No `version:` input exists anywhere in the workflows now, so both `TRIVY_IMAGE` pins were unmanaged — they would have aged exactly like the action default they replaced, which is the problem this PR set out to solve.

It now reads the image line and captures the digest as well:

```
"# renovate: datasource=(?<datasource>\\S+) depName=(?<depName>\\S+)\\s+[A-Z_]+: \"[^\":]+:(?<currentValue>[^\"@]+)(?:@(?<currentDigest>sha256:[a-f0-9]+))?\""
```

Tested against the two files, extracting all four fields:

```
.github/workflows/scan.yml         datasource=docker depName=aquasec/trivy
                                   currentValue=0.74.0 currentDigest=sha256:62b1e65e…
.github/workflows/verifyimage.yml  datasource=docker depName=aquasec/trivy
                                   currentValue=0.74.0 currentDigest=sha256:62b1e65e…
```

`renovate-config-validator` passes on the config. The digest group is optional, so a tag-only pin would still be picked up if one is ever added.

PR #460, opened 2026-08-16
fix(images): install runtime libraries instead of development packages

fzipi:
Follow-up to #459, which surfaced this: the largest share of the vulnerabilities reported against the debian images comes from packages the runtime stages never needed.

## What was wrong

`liblua${LUA_VERSION}` is a **virtual** package name in Debian. apt satisfies it with `liblua5.3-dev` (and `liblua5.3-0-dbg`), which depends on `libc6-dev`, which depends on `linux-libc-dev` — kernel headers, in the image, matched against every kernel CVE. The concrete runtime package is `liblua5.3-0`. Both `nginx/Dockerfile` and `apache/Dockerfile` were affected.

On top of that the nginx images asked for development packages directly: `libcurl4-gnutls-dev`, `libmaxminddb-dev` (debian) and `curl-dev`, `libmaxminddb-dev`, `libxml2-dev`, `lmdb-dev` (alpine). ModSecurity, the nginx modules and the Lua modules are compiled in the build stages, which install their own build dependencies, so the runtime stages only need the libraries.

## Effect

Same base image, only the package list changed:

| | before | after |
|---|---|---|
| packages | 218 | 170 |
| image size | 297MB | 231MB |
| vulnerabilities (all severities) | 1108 | 337 |
| HIGH + CRITICAL | 130 | 62 |
| fixable HIGH + CRITICAL | 0 | 0 |

The alpine images lose every `-dev` package they carried; `owasp/modsecurity-crs:nginx-alpine` scans clean.

## Verification

Built `nginx-debian-writable-latest`, `nginx-alpine-writable-latest` and `apache-debian-latest` locally and ran each container against a dummy backend:

- a path traversal request (`/?test=../../etc/passwd`) is blocked with `403` on all three, so ModSecurity loads and the rules run
- a normal request reaches the proxy (`502` from nginx, `503` from apache, both dummy-backend responses)
- `ldd /usr/local/modsecurity/lib/libmodsecurity.so.3` resolves every library in each image
- no `-dev` package, `libc6-dev` or `linux-libc-dev` remains in any image

`apache/Dockerfile-alpine` needed no change — its runtime stage already installs only runtime packages.

## Summary by CodeRabbit

* **Chores**
  * Updated container runtime packages to use runtime-only dependencies instead of development packages.
  * Improved image efficiency by removing unnecessary build tools and headers from production images.
  * Maintained required functionality for cURL, Lua, XML, database, security, and related runtime libraries across supported container variants.


theseion:
are the test failures related to the change?

fzipi:
No, we are still carrying those from before. This was added to try to fix it: https://github.com/coreruleset/modsecurity-crs-docker/pull/456, let me set that to 1.

theseion (approved):

PR #468, opened 2026-08-18
docs: stop advertising the read-only image variants

fzipi:
Refs #172.

The README promises images that are not published. `docker-bake.hcl` has the read-only matrix entry commented out:

```hcl
    read-only-fs = [
        {
            name = "writable"
            read-only = "false"
        },
        # {
        #     name = "read-only"
        #     read-only = "true"
        # }
    ]
```

so no read-only target is ever built or pushed. Confirmed against both registries:

```
ghcr:nginx-read-only          not found
hub:nginx-read-only           not found
ghcr:nginx-alpine-read-only   not found
hub:nginx-alpine-read-only    not found
```

The section now says so and points at #172, while noting that the machinery is still in the tree — the `READ_ONLY_FS` build argument in `nginx/Dockerfile` and `nginx/docker-entrypoint.d/0-move-writables.sh` — so re-enabling it later is a bake change rather than a rewrite.

The tag format sections had the same problem from the other direction: they document a `[-<writable>]` component, but the writable variant adds no suffix and the read-only one is not built, so no published tag has ever carried it. `owasp/modsecurity-crs:nginx-writable` does not resolve either. Removed from both the stable and rolling tag descriptions.

This documents the current state; whether the variants come back is the decision still open in #172.

## Summary by CodeRabbit

* **Documentation**
  * Updated stable and rolling tag documentation to remove the optional writable suffix.
  * Clarified that read-only filesystem variants are unpublished and do not produce pushed tags.
  * Documented that nginx build support for read-only variants remains available but disabled.
  * Removed the link to issue `#172` from the read-only filesystem documentation.


README.md
```diff
@@ -74,13 +72,14 @@ Examples:
...
-   * `nginx-read-only`
-   * `nginx-alpine-read-only`
+> Read-only filesystem variants are not currently published. The `nginx-read-only` and
+> `nginx-alpine-read-only` tags do not exist on Docker Hub or ghcr.io.
+
+The root filesystem of our containers is writable. Support for running on a read-only
+filesystem is still in the nginx images — the `READ_ONLY_FS` build argument and the
+`0-move-writables.sh` entrypoint script — but the build targets that would produce those
+images are disabled in `docker-bake.hcl`, so no such tags are pushed. See
+[#172](https://github.com/coreruleset/modsecurity-crs-docker/issues/172).
```
theseion: I don't think we need to reference that issue in the readme.

theseion (changes requested):

README.md
fzipi: Dropped the link in 997aa01.

theseion (approved):

Issue #385, opened 2025-10-01
Basic modsecurity setup with nginx docker image responds with HTML content but with content-type of text/plain


jpierson-at-riis:
When testing out the basic features of modsecurity, I tried accessing a non-existing page which is included by threat scanning tools to see how modsecurity would handle it. I was happy to see that the request resulted in a generic 403 response but instead of the HTML rendering in the browser it was displayed in plain text. When checking, it looks like the content-type of the response was `text/plain` when it really should have been `text/html`.

Additionally, the `server` header in the response was disclosing that the underlying system was using `nginx` which is a concern due to disclosure of the underlying technology, and which allows attackers narrow their possible set of attacks. In the same way, the default 403 error message show below also references `nginx`. My recommendation, if possible, would be to configure this product to remove these headers and to use something generic which does not signal anything unique to clue an attacker in on the use of modsecurity or nginx.

Image: `owasp/modsecurity-crs:nginx`
Example URL: `http://localhost:8080/.env`

Response: 
```html
<html>
<head><title>403 Forbidden</title></head>
<body>
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx</center>
</body>
</html>
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
```

I had brought this up originally at https://github.com/owasp-modsecurity/ModSecurity-nginx/issues/362#issuecomment-3352888117.

zyro23:
fwiw, looks like thats the default nginx error page which renders "nginx" even if `server_tokens off` is set. ref.:
* https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens
* https://github.com/nginx/nginx/blob/release-1.29.2/src/http/ngx_http_special_response.c

regarding the error response `Content-Type` header, it seems that is affected by the modsecurity-crs CORS config which does not look intentional, ref.:
* https://github.com/coreruleset/modsecurity-crs-docker/blob/release/20251003/nginx/templates/includes/cors.conf.template#L1

partial workaround in the meantime: set the environment variable `CORS_HEADER_403_CONTENT_TYPE` to `text/html`.

[...]

PR #469, opened 2026-08-18
feat(nginx): make the Server response header configurable

fzipi:
Addresses the second half of #385. The first half, the `Content-Type` of blocked responses, was fixed in #458.

`SERVER_TOKENS=off` is already the default, but it only decides how much nginx reveals about itself, never whether it identifies itself at all — `off` still answers `Server: nginx`, which is what the reporter saw. There was no way to change that without replacing the template.

`SERVER_HEADER` now sets the value, defaulting to `nginx` so nothing changes for existing users:

| `SERVER_HEADER` | `Server` header |
|---|---|
| `nginx` (default) | `nginx` |
| any string | that string |
| empty | header removed |

Measured on 403, 502 and 200 responses:

```
default (nginx)     403:[nginx] 502:[nginx] 200:[nginx]
custom (waf)        403:[waf]   502:[waf]   200:[waf]
empty -> dropped    403:[]      502:[]      200:[]
```

## Notes

The status codes are enumerated in the directive because headers-more rejects ranges — `-s '400-599'` makes nginx refuse to start — and without a list it covers only a handful of default codes, which left `Server: nginx` intact on a 502 in testing.

The two settings work at different layers, and the README now says so:

```
SERVER_TOKENS=on + SERVER_HEADER=waf
  header: Server: waf
  body:   <hr><center>nginx/1.30.4</center>
```

`SERVER_HEADER` wins for the header, because headers-more runs in the header filter after the response is built, while `SERVER_TOKENS` still governs the footer of nginx's own error pages.

That footer is the one part of #385 this does not address: nginx hardcodes its name in the built-in error page body, and only an `error_page` with a bundled file removes it. That changes the visible 403 for every user and interacts with the CORS `Content-Type` we set on 403 responses, so I left it out and would rather do it deliberately if wanted. #385 should stay open for that.

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing the HTTP `Server` response header through the `SERVER_HEADER` environment variable.
  * Applied the default `nginx` server header across standard and Alpine container images.
  * Header customization supports explicitly specified HTTP status codes.

* **Documentation**
  * Clarified how `SERVER_HEADER` and `SERVER_TOKENS` affect server identification and error-page details.


fzipi:
Three CodeRabbit findings, two applied in 11cb440 and one skipped with evidence.

**1. Status list omitted codes — valid, and the suggested fix was right.** I had claimed `-s` was necessary; that came from a probe with the directive inside a `location` block rather than at `http` level, which is not comparable. The [module documentation](https://github.com/openresty/headers-more-nginx-module) is explicit: *"Unlike the standard headers module, this module's directives will by default apply to all the status codes, including 4xx and 5xx."* Measured with probe locations returning 202, 416 and 511:

```
current list    200=[waf] 403=[waf] 202=[nginx] 416=[nginx] 511=[nginx] 502=[waf]
no -s at all    200=[waf] 403=[waf] 202=[waf]   416=[waf]   511=[waf]   502=[waf]
```

The enumeration was both unnecessary and incomplete, so the directive is now simply:

```nginx
more_set_headers 'Server: ${SERVER_HEADER}';
```

Re-verified across all three values:

```
SERVER_HEADER=nginx    200=[nginx] 403=[nginx] 202=[nginx] 416=[nginx] 511=[nginx] 502=[nginx]
SERVER_HEADER=waf      200=[waf]   403=[waf]   202=[waf]   416=[waf]   511=[waf]   502=[waf]
SERVER_HEADER=<empty>  200=[none]  403=[none]  202=[none]  416=[none]  511=[none]  502=[none]
```

**2. Escaping or validating the value — measured, not adding code.** The two failure modes both stop the container at startup with a precise message:

| value | result |
|---|---|
| `waf` | running, `Server: waf` |
| `Acme's WAF` | refused: `unexpected "s" in /etc/nginx/conf.d/default.conf:9` |
| `$hostname` | running, `Server: 25d82aa384ba` |
| `$nope_undefined` | refused: `unknown "nope_undefined" variable` |

An operator setting a malformed value learns immediately, and nothing starts serving with a wrong header, so escaping logic in an entrypoint script would only replace one clear error with another. Variable expansion is deliberate and matches `NGINX_X_FORWARDED_PORT='$server_port'` and `PROXY_HOST_HEADER='$host'`, which are documented the same way — `$hostname` is a useful value here, not a bug. The README now states that the value lands in an nginx directive, that `$name` is read as a variable, and that a quote in the value stops nginx from starting.

**3. `SERVER_TOKENS` described as a boolean — fixed.** Now reads "Controls how much server-identifying information is emitted", with the hyphen.

theseion (approved):

